PR #2730, opened 2015-01-05
Removed "Button" and replaced with "Filter"

DavidShaughnessy:


gregmoser:
These button text's needs to be RB Keys

PR #2743, opened 2015-01-09
Adds currency to no qualifier promotion calculations

dzikakis:
Would like to discuss the modified functions on the sku and product entities. I've added the currencyCode argument to the existing functions rather than duplicating these to specific "byCurrencyCode" functions. I know this goes against the established convention on these entities, but it seemed redundant. Really didn't want to duplicate the query in the DAO for maintenance reasons.

Things to note: As-is amount off and fixed amount promotions will take the same amount off (number wise) regardless of currency. Not sure which would be desired behavior here - exclude these two types, include qualifying currencies, keep as is?

### DIFF
--- a/model/dao/PromotionDAO.cfc
+++ b/model/dao/PromotionDAO.cfc
@@ -304,7 +304,8 @@ Notes:
 	
 	<cffunction name="getSalePricePromotionRewardsQuery">
 		<cfargument name="productID" type="string">
-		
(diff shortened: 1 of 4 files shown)


gregmoser:
Need currencyCode on SwSku table
Need currencyCode on SwPromotionReward table
Sku currency conversion need to take into account old currency vs new currency

gregmoser:
Closing, please re-issue new request once updates made

PR #2745, opened 2015-01-09
removed the ?slatAction= from the value of the action parameter

christopher-kundrat:


gregmoser:
This should actually be:

```
getFW().redirect(action="entity.detailaccount", queryString="accountID=#arguments.rc.accountid#", preserve="messages");
```

PR #2758, opened 2015-01-14
Develop sa public ui

webdevsourcerer:
Updates to Files for StandAlone Front End GUI


gregmoser:
Can you give some more insights as to the intent of this pull request?  It looks like there are a number of things in here...

I also want to make sure that you are aware of the /meta/sample application that is available.

webdevsourcerer:
Hey Greg,
This is for the 3.x Stand-Alone Front-end UI. Its Ultra Basic, but if you prefer I can Flesh it out some more and keep developing my SA-Public-UI Build.

No I was not aware of the /meta/sample aplication.

gregmoser:
Why don't you keep developing on your branch... I also encourage your to take a look at the sample app as that has an existing frontend out of the box that you can use.  If there is still a need for some of this code after you review our sample frontend we can talk about it.

PR #2762, opened 2015-01-16
Feature rm orderpayment

rmarchandTen24:
adding the ability to select from and use previous order payment methods


rmarchandTen24:
making language for process object properties more distinct. Adding active filter on the get previous orderPayments smartlist

PR #2766, opened 2015-01-19
Hotfix

gregmoser:


gregmoser:
Just waiting on the additional ckeditor files before this gets merged into master

PR #2768, opened 2015-01-19
Removed variable displaying boolean value in HibachiListingDisplay

DavidShaughnessy:
Removed the boolean being displayed.  There didn't seem to be a need to display true/false to end user. This was not in Master either.


gregmoser:
Looks like somehow it came from a combination of these two commits:

https://github.com/ten24/Slatwall/commit/541f9ab9eb05bae3969b9e887b9eab85b4fd027b
https://github.com/ten24/Slatwall/commit/1af8cb05e3b7fb0df7edcfda377f02dfc3731cfd

PR #2769, opened 2015-01-19
Fixed angularUIBootstrap showweeks in datepickerConfig to default false

DavidShaughnessy:


rmarchandTen24:
Hi Dave, we should never have to edit the library file itself. True that it is defaulted to show weeks, however you should be able to find a way to override the default. So we include the datepicker into angular via a module. therefore you should be able to configure the module. An example of how to do that can be found here http://plnkr.co/edit/6i4G7JkvBiWXZYlrV2GL?p=preview. 

Tasks:
1. revert ui.bootstrap.js
2. add configuration to slatwall admin

PR #2773, opened 2015-01-20
Exclude the current uncommitted order payment from the total by payment method

dzikakis:
Small update to exclude the current uncommitted order payment from the total by payment method.


rmarchandTen24:
please recreate this pull request into the feature branch instead of develop

PR #2775, opened 2015-01-20
Addressed ryan's comments and fixed at the module configuration level

DavidShaughnessy:


rmarchandTen24:
Hi Dave, this is good, however you don't need the ng-controller. Please remove the ngcontroller.

rmarchandTen24:
hi Dave, this still needs the removal of the  ng-controller="DatepickerCtrl" from criteriadate.html.

can you make that change and push it?

PR #2777, opened 2015-01-20
Hotfix orbital status bug

christopher-kundrat:


integrationServices/orbital/Payment.cfc
```diff
@@ -193,7 +193,10 @@ component accessors="true" output="false" displayname="PayFlowPro" implements="S
 			// Transaction did not go through
 			response.addError(responseData.ProcStatus.xmlText, responseData.StatusMsg.xmlText);
 		} else if(structKeyExists(responseData,"ApprovalStatus") && responseData.ApprovalStatus.xmlText != 1) {
-			// Transaction was not approved
+			if (structKeyExists(responseData, "RespCode") && structKeyExists(responseData.RespCode, "xmlText")) {
+				// Transaction was not approved
+				response.setStatusCode( responseData.RespCode.xmlText );
+			}
 			response.addError(responseData.RespCode.xmlText, responseData.StatusMsg.xmlText);
```
sumitverma: If you are checking the struct key then the next line should be inside the conditional as well. And you will need to add an else condition with some custom message that transaction was declined (rb key).

gregmoser: agreed

PR #2778, opened 2015-01-21
Feature ds click outside directive

DavidShaughnessy:
-Added Directive to handle watching for click events
-Added callback function in controller that expects a struct of arguments
-Added callback arguments to productBundleGroupTypes and implemented directive


DavidShaughnessy:
Comments addressed and functionality added to PR to come.

PR #2781, opened 2015-01-21
Feature ds product bundle add screen

DavidShaughnessy:
-Added swclickoutside directive to watch for click event targets outside of designated area
-Added callback to be passed into swclickoutside directive and executed
-Added close function 
-Added clear typeName function
-Fixed button font awesome carets and added ngShow conditional to display between the two 
-Added Ryan's change to _save functionality in ngslatwall so that timeout promise is returned


rmarchandTen24:
Hi Dave, we shouldn't be dependent j-inside and j-outside attributes. this would prevent us from having multiple of these items because id is exclusive.

we should be using the e.target and element for comparison.

PR #2801, opened 2015-01-26
Fixed up some rbkeys

DavidShaughnessy:


admin/client/js/directives/swproductbundlegrouptype.js
```diff
@@ -46,7 +46,7 @@ angular.module('slatwalladmin')
 					productBundleGroupType.$$setParentType(parentType);
 					productBundleGroupType.data.typeName=$scope.productBundleGroup.data.productBundleGroupType.data.typeName;
 
-					productBundleGroupType.data.typeDescription = '';
+					productBundleGroupType.data.description = '';
```
gregmoser: I do not think that changing the actual data field is correct.  the field in the database is still typeDescription, we just want the RBKey to be 'Description'.  Ryan, can verify if this is correct or not.

DavidShaughnessy: No you're right.  I'll revert it now.

PR #2805, opened 2015-01-26
Removed incrementation and added it back in upon saving of filter

DavidShaughnessy:


rmarchandTen24:
the following code increments the count by 1 everytime the save button is clicked. We should be checking/maintaining the filteritem count.

rmarchandTen24:
I think that the places where the incrementing happens was in the right place originally but needs to only increment when the filter item is persisted. I think that the filter items have a property of $$isNew() somewhere. If we check that property before running the increment function, you should be in good shape.

DavidShaughnessy:
Ok I'll make the change

On Tue, Jan 27, 2015 at 1:31 PM, rmarchandTen24 <notifications@github.com>
wrote:

> I think that the places where the incrementing happens was in the right
> place originally but needs to only increment when the filter item is
> persisted. I think that the filter items have a property of $$isNew()
> somewhere. If we check that property before running the increment function,
> you should be in good shape.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/ten24/Slatwall/pull/2805#issuecomment-71732035>.
>



-- 
*David Shaughnessy*
Application Developer | ten24 Web Solutions
http://www.ten24web.com

PR #2808, opened 2015-01-27
Added $apply() to clickoutside directive and fixed bundgrouptype save

DavidShaughnessy:


rmarchandTen24:
Hey Dave, 
we should never actually use $apply() in the application. In the instances where it may be necessary it would be better to use $timeout because in runs an apply but only after all other digest cycles have finished.

What is the reason that you need to use apply in this case?

PR #2809, opened 2015-01-27
clean up design and add new search with loader

Reyjay:
Fix small structure issues of collections.  
Add new search input design.  
Add loading wheel


gregmoser:
Whats the next step on this guy?  Are we going to merge it in, or not merge it?  If we need to merge it in, then it needs to be updated with the most recent version of 'feature' first.

PR #2819, opened 2015-01-28
Feature show estimated fulfillment date in order view

christopher-kundrat:


model/entity/Order.cfc
```diff
@@ -441,6 +445,50 @@ component displayname="Order" entityname="SlatwallOrder" table="SwOrder" persist
 		return fulfillmentChargeAfterDiscountTotal;
 	}
 	
+	public any function getNextFulfillmentDateTime(){
+    	if(arrayLen(getOrderFulfillments())) {
+    		var nextFulfillmentDateTime = '';
+    		
+    		//Loop over orderFulfillments to get the nextFulfillmentDateTime
+			for(var i=1; i<=arrayLen(getOrderFulfillments()); i++){
```
gregmoser: for(var orderFulfillment in getOrderFulfillments()) {}

model/entity/Order.cfc
```diff
@@ -441,6 +445,50 @@ component displayname="Order" entityname="SlatwallOrder" table="SwOrder" persist
 		return fulfillmentChargeAfterDiscountTotal;
 	}
 	
+	public any function getNextFulfillmentDateTime(){
+    	if(arrayLen(getOrderFulfillments())) {
+    		var nextFulfillmentDateTime = '';
```
gregmoser: set this value as getEstimatedFulfillmentDateTime()

gregmoser:
- Change to "NextESTIMATED"
- Add persistent properties at the orderFulfillment level
- Add task to do example inputs on frontend
- change the "Next" loop logic to just return earliest 'undelivered' orderItem estimate... that orderItem estimate may come from item, fulfillment or order.

PR #2820, opened 2015-01-28
Feature show estimated fulfillment date in order view

christopher-kundrat:


config/resourceBundles/en.properties
```diff
@@ -2238,6 +2238,10 @@ entity.order.calculatedTotal=Total
 entity.order.cancel_validatestatus=Order cannot be deleted due to order status ({statusValue})
 entity.order.defaultStockLocation=Default Stock Location
 entity.order.discountTotal=Discount
+entity.order.estimatedFulfillmentDateTime=Estimated Order Fulfillment Date Time
+entity.order.estimatedDeliveryDateTime=Estimated Order Delivery Date time
+entity.order.nextFulfillmentDateTime=Next Order Fulfillment Date Time
+entity.order.nextDeliveryDateTime=Next Order Delivery Date time
```
gregmoser: These rbKeys can just be entity.define.xxx and the 'next' ones will need to be updated to 'nextEstimated' as the properties have changed

PR #2824, opened 2015-01-29
New solutions that grabs the true count of filters & sets vs increments

DavidShaughnessy:


rmarchandTen24:
I see that there is break out but i think you need an everlong in your for loop

PR #2827, opened 2015-01-30
Add remoteID to PriceGroup

chriskent:


gregmoser:
Will you just re-create this pull request into 'hotfix' instead of 'master'

PR #2834, opened 2015-02-02
Merge branch 'hotfix' into hotfix-account-last-cart

gregmoser:


gregmoser:
For some reason I thought that this had already been merged into master, but I wanted to create it here as I'm now not sure that it ever did.  We need to research the teamwork task

PR #2835, opened 2015-02-02
Pointed path to custom/config/model directory

DavidShaughnessy:
This was an issue I ran into while working on a separate task.  custom/config/model/entity does not exist in the file system


gregmoser:
Then the solution is to create that folder by default in Slatwall with a readme.md

gregmoser:
I opened this PR: https://github.com/ten24/slatwall/pull/2836

PR #2844, opened 2015-02-03
Fixed a validation error that was busting up the ability to add skus

DavidShaughnessy:


gregmoser:
This validation is inaccurate, we shouldn't be allowing for "Add Option Group" "Add Option" or "Add Sku" if it is an event product.

PR #2849, opened 2015-02-04
Fixed issue.  Was calling wrong function in collectionService

DavidShaughnessy:


rmarchandTen24:
there is still an issue when creating a filter group and then clicking save add another. It saves and then breaks the collection

PR #2855, opened 2015-02-05
changed maintainQueryString to false

DavidShaughnessy:
Changed boolean value because if delete is successful we should not be maintaining the original query string as the primaryID will have been deleted.


gregmoser:
This will break a bunch of stuff.  It cannot be merged in.

rmarchandTen24:
there is still an issue when creating a filter group and then clicking save add another

rmarchandTen24:
need to review this with greg

PR #2871, opened 2015-02-09
Corrected case for sw:formField to sw:FormField in AddressForm.cfm

chriskent:
Corrected case for sw:formField to sw:FormField in AddressForm.cfm


gregmoser:
This looks good

PR #2893, opened 2015-02-12
Hotfix admin order sku stock tabs

chriskent:
Add new attribute table ids, to allow multiple occurrences of same entity listing on a page
And add tabid's for sale order items and return order items

### DIFF
--- a/admin/views/entity/ordertabs/addsku.cfm
+++ b/admin/views/entity/ordertabs/addsku.cfm
@@ -54,6 +54,7 @@ Notes:
 
 <cfoutput>
 	<hb:HibachiListingDisplay smartList="#rc.order.getAddOrderItemSkuOptionsSmartList()#"
+							  tableID="#rc.addSkuAddStockType#"
 							  recordProcessAction="admin:entity.processOrder"
 							  recordProcessQueryString="orderItemTypeSystemCode=#rc.addSkuAddStockType#"
 							  recordProcessContext="addOrderItem"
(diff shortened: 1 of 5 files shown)


gregmoser:
I'm not comfortable having this merged in as it doesn't explain why these custom ID's need to be generated in the first place.  The tables in the listing displays use the getSavedStateID() function to return a UUID that is SUPPOSED to be unique to that smartList cfc.  The question is, why would the orderItem and returnOrderItem smart list have the same ID in the first place.

chriskent:
The issue is that for exchange orders, there are 2 tabs, one for Sale Order Items and a second for Return Order Items.

Both of these tabs include HibachiListingDisplay to Add SKU and Add Stock using ordertabs/addsku and ordertabs/addstock. 

These use rc.order.getAddOrderItemSkuOptionsSmartList() and rc.order.getAddOrderItemStockOptionsSmartList() to build the lists in HibachiListingDisplay 

So HibachiListingDisplay is being used twice in the admin page request for each smart list. This builds the html tables twice with the same ID from attributes.smartList.getSavedStateID(), The duplication of the ID us what is causing the issue with duplicate columns of data on pagination. 

The second issue fixed as to assign unique id's to the Add SKU and Add Stock tabs, as these where only working on the Sale Order Item tab and not the Return Order Item tab.

gregmoser:
Ahh, I see.  Ok, we can go over this on our call today.

gregmoser:
Based on our conversation... Chris is going to re-issue this PR with a slightly different solution to the same issue.

PR #2895, opened 2015-02-13
Remove duplicate code blocks from orderservice.cfc

chriskent:


gregmoser:
Looks good to be merged, however this needs to be issued into 'feature' instead of develop.

PR #2897, opened 2015-02-15
Fix error when calling the process object function

christopher-kundrat:


gregmoser:
This is a good fix and should be merged.  Without the ```this.``` onMissingMethod doesn't work.

PR #2904, opened 2015-02-16
Feature ds bundle group selection search alpha

DavidShaughnessy:


admin/client/js/directives/swproductbundlegroup.js
```diff
@@ -146,7 +147,19 @@ angular.module('slatwalladmin')
...
+											        }else{
+											            return 0;   
+											        }
+												});
+												$log.debug(scope.productBundleGroupFilters.value);
+												scope.loading = false;
+												
 											}
 										});
 									})(keyword,option);
```
rmarchandTen24: on the else statement we need that part alphabetized too

PR #2906, opened 2015-02-16
Make Cursor a pointer on filter

Reyjay:
Need to make the filter remove icon a pointer so people know it is clickable 

Task:https://ten24.teamworkpm.net/tasks/3151397


rmarchandTen24:
please update and resissue

PR #2907, opened 2015-02-16
Feature rj filter title 3161421

Reyjay:
Fixed filter title to work with any length and to not cut off comma.

Task: https://ten24.teamworkpm.net/tasks/3161421


Reyjay:
Add an angular toggle to the style guide as an example.

PR #2909, opened 2015-02-16
Allow for multiple occurrences of getAddOrderItemSkuOptionsSmartList & g...

chriskent:
Allow for multiple occurrences of getAddOrderItemSkuOptionsSmartList & getAddOrderItemStockOptionsSmartList for sales order items and return orders items
Also fix for Add SKU and Add Stock for sales order items and return orders items

### DIFF
--- a/model/entity/Order.cfc
+++ b/model/entity/Order.cfc
@@ -289,24 +289,20 @@ component displayname="Order" entityname="SlatwallOrder" table="SwOrder" persist
 	// ============ START: Non-Persistent Property Methods =================
 	
 	public any function getAddOrderItemSkuOptionsSmartList() {
-		if(!structKeyExists(variables, "addOrderItemSkuOptionsSmartList")) {
-			variables.addOrderItemSkuOptionsSmartList = getService("skuService").getSkuSmartList();
-			variables.addOrderItemSkuOptionsSmartList.addFilter('activeFlag', 1);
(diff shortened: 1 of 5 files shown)


model/entity/Order.cfc
```diff
@@ -289,24 +289,20 @@ component displayname="Order" entityname="SlatwallOrder" table="SwOrder" persist
 	// ============ START: Non-Persistent Property Methods =================
 	
 	public any function getAddOrderItemSkuOptionsSmartList() {
-		if(!structKeyExists(variables, "addOrderItemSkuOptionsSmartList")) {
-			variables.addOrderItemSkuOptionsSmartList = getService("skuService").getSkuSmartList();
-			variables.addOrderItemSkuOptionsSmartList.addFilter('activeFlag', 1);
-			variables.addOrderItemSkuOptionsSmartList.addFilter('product.activeFlag', 1);
-			variables.addOrderItemSkuOptionsSmartList.joinRelatedProperty('SlatwallProduct', 'productType', 'inner');
-			variables.addOrderItemSkuOptionsSmartList.joinRelatedProperty('SlatwallProduct', 'brand', 'left');
-		}
```
gregmoser: Instead of using variables.addOrderItemSkuOptionsSmartList... just use something like 

```
var optionsSmartList = ...
return optionsSmartList;
```

sense we aren't caching it, we don't need to ever stick it in the variables scope of this object.

PR #2915, opened 2015-02-17
If paywall content is set to display=no, redirect to 404 page instead of...

chriskent:
If paywall content is set to display=no, redirect to 404 page instead of displaying content or barrier page

Reported by Scientific American

### DIFF
--- a/integrationServices/mura/model/handler/MuraEventHandler.cfc
+++ b/integrationServices/mura/model/handler/MuraEventHandler.cfc
@@ -369,10 +369,18 @@
 				$.slatwall.setContent( barrierPage );
 				
 				// Update the mura content to use the barrier page or 404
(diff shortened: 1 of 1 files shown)


gregmoser:
This looks good.  I'm going to forward the PR to Nick and make sure it's what he is looking for before we merge it in.

nsollecito:
IMO it might be more appropriate to allow Mura to handle the 404 through its event model in the case of existing custom handlers. 

With this change you're forcing a redirect where something like $.announceEvent('standard404Handler') (or maybe it's $.announceEvent('onSite404')) might be more appropriate?

gregmoser:
I was a bit concerned with that as well.  I believe that if the original content is set to display 'no' we can fundamentally "do nothing" and it should work correctly.  Chris, can you give that a shot and let us know if doing nothing actually works?

gregmoser:
Closing this for now until Chris can resolve the concerns brought up and issue a new PR

PR #2917, opened 2015-02-18
New more directed fix for the delete redirect issue

DavidShaughnessy:


gregmoser:
Hey david, this is 99% of the way there... however instead of delegating "itemEntityName" can we change the argument that gets passed to be "keysToRemoveOnRedirect" and then from the delete() function call:

```
renderOrRedirectSuccess( defaultAction=arguments.rc.entityActionDetails.listAction, maintainQueryString=true, rc=arguments.rc, keysToRemoveOnRedirect=entityPrimaryID);
```

The value of "entityPrimaryID" is set at the top of the delete function and this accounts for things like ```countryCode``` so that is isn't done as ```#arguments.rc.entityActionDetails.itemEntityName#ID"``` which would have resulted in ```countryID``` which in that case would be wrong

org/Hibachi/HibachiControllerEntity.cfc
```diff
@@ -698,11 +702,11 @@ component output="false" accessors="true" extends="HibachiController" {
...
 	}
 	
-	private string function buildRedirectQueryString( required string queryString, required boolean maintainQueryString, struct rc ) {
+	private string function buildRedirectQueryString( required string queryString, required boolean maintainQueryString, string itemEntityName ) {
 		if(arguments.maintainQueryString) {
 			for(var key in url) {
-				if(isDefined('arguments.rc.entityActionDetails.itemEntityName')){
-					if(key != getFW().getAction() && !listFindNoCase("redirectAction,sRedirectAction,fRedirectAction,redirectURL,sRedirectURL,fRedirectURL,#arguments.rc.entityActionDetails.itemEntityName#ID", key)) {
+				if(isDefined('arguments.itemEntityName')){
+					if(key != getFW().getAction() && !listFindNoCase("redirectAction,sRedirectAction,fRedirectAction,redirectURL,sRedirectURL,fRedirectURL,#arguments.itemEntityName#ID", key)) {
```
gregmoser: Also, this linke will need to change to:
```
if(key != getFW().getAction() && !listFindNoCase("redirectAction,sRedirectAction,fRedirectAction,redirectURL,sRedirectURL,fRedirectURL,#arguments.keysToRemoveOnRedirect#", key)) {
```

gregmoser:
Ok, 1 more thing... I'll ping you directly via IM to go over the change.

gregmoser:
All good on this now.

PR #2920, opened 2015-02-19
Vertex & Core tax updates

gregmoser:
Awate confirmation for SciAm before merge


gregmoser:
This is good to merge now, please include in the next hotfix release.

PR #2925, opened 2015-02-19
Added a property in Attributes 

ian-a-hickey:
(A flag for displaying attributes on the order item admin).
Added the flag on the basic tab of the attributes display.
Added an RBKey for the display.


gregmoser:
I think the text of "Display In Order Item Admin" is a bit confusing.  Could we change this to something like "Display on order detail page"... and maybe make the setting "displayOnOrderDetailFlag"

ian-a-hickey:
No problem.

admin/views/entity/attributetabs/basic.cfm
```diff
@@ -8,6 +8,7 @@
 	<hb:HibachiPropertyRow>
 		<hb:HibachiPropertyList>
 			<hb:HibachiPropertyDisplay object="#rc.attribute#" property="activeFlag" edit="#rc.edit#">
+			<hb:HibachiPropertyDisplay object="#rc.attribute#" property="displayOnOrderDetailFlag" edit="#rc.edit#">
```
gregmoser: This needs a conditional around it and it should only be when the attributeSet is an "Order Item" attribute set.

```
<cfif rc.attributeSet.getAttributeSetObject() == "OrderItem">
... property display ...
</cfif>
```

PR #2947, opened 2015-02-24
Added replace to add space between commas in displayValue of filteritem

DavidShaughnessy:


rmarchandTen24:
can we handle this during the adding method rather than afterwards?

PR #2948, opened 2015-02-24
Feature rj form validation 3216749 

Reyjay:
Added a default form validation to angular structured forms with a fallback for old style forms.

Task: https://ten24.teamworkpm.net/tasks/3216749


rmarchandTen24:
needs conflict resolution

PR #2950, opened 2015-02-24
Removed delimiter so that the entire string would be used to search with

DavidShaughnessy:


gregmoser:
This is not a valid solution.  Each word needs to get searched independently.  This will search the words in a specific order which is not what we want.  The words should still be split up by space, but the filters added should be "AND" instead of "OR"

PR #2954, opened 2015-02-25
Update basic.cfm

dzikakis:
Was erroring out when creating a new attribute, because rc.attribute.attributeSet wasn't populated yet.


rmarchandTen24:
fixed with
!isNull(rc.attributeSet.getAttributeSetObject()) and rc.attributeSet.getAttributeSetObject() EQ "OrderItem"

PR #2956, opened 2015-02-25
Created a displayValue at the criteria directive level

DavidShaughnessy:


rmarchandTen24:
This was reverted. The space not be added in the watch method, but instead the add method. Also the commas are not being removed when deleting the items

PR #2957, opened 2015-02-25
beefing up stock smartlist 

dzikakis:
Adding in joins on sock smartlist similar to sku smartlist so things are searchable in the admin/order/order items/stock tab


gregmoser:
This is good to merge, thanks David

PR #2960, opened 2015-02-25
Fixes bug of incorrect return quantity being returned

dzikakis:


gregmoser:
Good to Merge

gregmoser:
Thanks for this David, it's a great catch.

PR #2976, opened 2015-03-02
Hotfix ih ckeditor span mystery

ian-a-hickey:


ian-a-hickey:
This branch is called 'hotfix' but I created it off of feature branch and should be merged into feature.

PR #2988, opened 2015-03-04
Update Payment.cfc

brucebailey:
So it appears Stripe changed their API in February and the attribute card and default_card and not longer in play and source and default_source should be used respectively.  I changed those up in my Payment.cfc reloaded and it worked like a champ


gregmoser:
Thanks for contributing bruce!

PR #2990, opened 2015-03-05
Re-added ability to add an orderitem (using original display). High Priority

ian-a-hickey:
Additionally - Fixed Attribute Set error that was causing orderitems to not be added by adding a null check.


DavidShaughnessy:
More work to be done!

PR #2995, opened 2015-03-06
Added rbkeys to orderitems administrator.

ian-a-hickey:
Added new rbkeys.


gregmoser:
should be define.admin.orderdetail.xxx

ian-a-hickey:
fixed.

gregmoser:
sorry Ian, I told you the wrong thing... it should just be:

admin.orderdetail.xxx

stuff that starts with 'define' should be generic terms that would be used anywhere.  In fact a lot of this could be done as just define... here is what I would do:

define.image=Image (probably already there)
define.title=Title (probably already there)
entity.sku=Sku (already there)
admin.orderdetail.orderitems.details=Details
admin.orderdetail.orderitems.customization=Customizations
define.price=Price (probably already there)
define.qty=QTY (probably already there)
define.discount=Discount (probably already there)
define.total=Total (probably already there)
define.tax=Tax (probably already there)
define.fulfillment=Fulfillment

At the end of the day we are trying to create as few rbKeys as possible.

ian-a-hickey:
Not a problem. I would rather know how to do it correctly.  Thanks.

PR #3003, opened 2015-03-09
moving prototypes

rmarchandTen24:
moving prototypes to the meta folder


gregmoser:
also we need to delete ngtest.cfm

PR #3008, opened 2015-03-10

DavidShaughnessy:


rmarchandTen24:
we shouldn't be doing a new request upon removing an item. Either the item should be pushed back into the list or the items on the left should never be spliced to being with and instead hidden

PR #3010, opened 2015-03-10

DavidShaughnessy:


rmarchandTen24:
still a few tweaks. Either run a re-alphabetize function after doing a push or convert to using a show/hide strategy

PR #3020, opened 2015-03-10

DavidShaughnessy:


rmarchandTen24:
can we make the alphabetize code into it's own function so we don't have to maintain it in multiple places?

PR #3023, opened 2015-03-10
Update script to set orderFulfillmentStatusTypeID

chriskent:


config/scripts/database/v4_0.cfm
```diff
@@ -118,6 +118,31 @@ Notes:
...
+				SELECT CASE
+					WHEN min(oi.quantity) - sum(odi.quantity) = 0 THEN '159118d67de3418d9951fc629688e194'
+					WHEN min(oi.quantity) - sum(odi.quantity) > 0 THEN 'fefc92c1d8184017aa65cdc882bdf636'
+					ELSE 'b718b6fadf084bdaa01e47f5cc1a8265'
+				END
+				FROM SwOrderItem oi
+					LEFT JOIN SwOrderDeliveryItem odi ON odi.orderItemID = oi.orderItemID
+				WHERE oi.orderFulfillmentID = SwOrderFulfillment.orderFulfillmentID
+       		)
+	</cfquery>
```
sumitverma: We need to make sure this query is compatible with mysql, sql server and oracle.

gregmoser:
Also this PR should be made into 'feature' I believe.  Ryan can confirm

config/scripts/database/v4_0.cfm
chriskent: I have tested against MS SQL, do not have Oracle installed and no current V4 MySQL database available.

dzikakis: I can confirm this query works on MySQL 5.5+

rmarchandTen24:
Confirmed

PR #3035, opened 2015-03-12

DavidShaughnessy:


rmarchandTen24:
to abstract the utilityService even more can you remove all the if statements and instead use

a[keyToSortBy] > a[keyToSortBy]

PR #3036, opened 2015-03-12
Fixes to vertex invoicing and currencyConversion

gregmoser:


gregmoser:
This has been tested via Scientific American and can be merged in.

PR #3037, opened 2015-03-12
Fix for caching engine for railo update.

gregmoser:
Hold off on merge until we confirm that this resolves the issue.


gregmoser:
Ok, this has been confirmed resolved and can be merged into the stable release.

PR #3046, opened 2015-03-16
Hotfix rm subscriptions payment

rmarchandTen24:
fixing bug where notifications are


model/service/SubscriptionService.cfc
```diff
@@ -443,11 +443,11 @@ component extends="HibachiService" persistent="false" accessors="true" output="f
 					
 				} else if (arguments.processObject.getRenewalPaymentType() eq 'orderPayment') {
 					var orderPayment = getOrderService().newOrderPayment();
-					
-					orderPayment.copyFromOrderPayment( arguments.processObject.getOrderPayment() );
+					if(!isnull(arguments.processObject.getOrderPayment())){
+						orderPayment.copyFromOrderPayment( arguments.processObject.getOrderPayment() );
+					}
```
sumitverma: If order payment is null, we should skip the whole block and not create a new order payment. Also, need to add a criteria here to check for AutoPay. If AutoPay is set to no, then no payment should be applied.

model/service/SubscriptionService.cfc
```diff
@@ -462,13 +462,13 @@ component extends="HibachiService" persistent="false" accessors="true" output="f
 				// Place the Order
 				order = getOrderService().processOrder(order, {}, 'placeOrder');
 				
-				// set the subscription usage nextBillDate
-				arguments.subscriptionUsage.setNextBillDate( nextBillDate );
-				arguments.subscriptionUsage.setFirstReminderEmailDateBasedOnNextBillDate();
-
```
sumitverma: Once the order is created nextBillDate and email date needs to be set irrespective of status because we don't want to run the renewal again. So, this needs to stay here.

model/service/SubscriptionService.cfc
```diff
@@ -462,13 +462,13 @@ component extends="HibachiService" persistent="false" accessors="true" output="f
 				// Place the Order
 				order = getOrderService().processOrder(order, {}, 'placeOrder');
```
sumitverma: If place order comes back with an error, we needs to set error in subscriptionUsage (processObject) as well, so the processSuccess event doesn't gets triggered. Also, if the order is not placed successfully, the order status needs to get changed to "New" (and new order number generated), so, it is tied to the renewal.

So, basically, the place order logic needs to be run even if payment is not applied/captured.

PR #3047, opened 2015-03-16
Fixes issue where return items sku and stock are added to the listing as

ian-a-hickey:
oitSale instead of oitReturn items.


rmarchandTen24:
needs review regarding getAddress method

PR #3049, opened 2015-03-16
Fixed the tabs

DavidShaughnessy:


gregmoser:
This doesn't look right.  Why is the EntityActionBar on there twice... including once with all of the details on the Basic Tab.

PR #3088, opened 2015-03-20
confirmation modal

Reyjay:
Add confirmation modal to style guide for refrence
https://ten24.teamworkpm.net/tasks/3354352


rmarchandTen24:
need conflict resolution

PR #3116, opened 2015-03-25
Fixed session DateTime issue using date instead.

ian-a-hickey:


org/Hibachi/HibachiSessionService.cfc
```diff
@@ -236,7 +236,7 @@ component output="false" accessors="true" extends="HibachiService"  {
 	 * Generate new cookie value
 	 */
 	private any function getValueForCookie(){
-		var id = getHibachiScope().getSession().getSessionID() & "#DateTimeFormat(Now())#" & RandRange(1, 100);
+		var id = getHibachiScope().getSession().getSessionID() & createUUID();
```
sumitverma: I think the idea here was to append some string _after_ hashing, because hash of 2 separate string can result in same value.

PR #3121, opened 2015-03-26
Update account.cfc

UltimoTG:
Send out verification email only if the email address has not been verified

### DIFF
--- a/public/controllers/account.cfc
+++ b/public/controllers/account.cfc
@@ -158,7 +158,7 @@ component output="false" accessors="true" extends="Slatwall.org.Hibachi.HibachiC
 		
 		var accountEmailAddress = getAccountService().getAccountEmailAddress( rc.accountEmailAddressID );
 		
-		if(!isNull(accountEmailAddress)) {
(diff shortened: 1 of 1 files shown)


public/controllers/account.cfc
```diff
@@ -158,7 +158,7 @@ component output="false" accessors="true" extends="Slatwall.org.Hibachi.HibachiC
 		
 		var accountEmailAddress = getAccountService().getAccountEmailAddress( rc.accountEmailAddressID );
 		
-		if(!isNull(accountEmailAddress)) {
+		if(!isNull(accountEmailAddress) and !accountEmailAddress.getVerifiedFlag()) {
```
gregmoser: should probably make sure it isn't null as well:

```
if(!isNull(accountEmailAddress) && !isNull(accountEmailAddress.getVerifiedFlag()) && !accountEmailAddress.getVerifiedFlag()) {
```

gregmoser:
Can you issue this Pull Request into 'hotfix' instead of master.  That way our automated tests will run once merged.

PR #3124, opened 2015-03-26
Feature ck order item promotion update

christopher-kundrat:


model/dao/PromotionDAO.cfc
```diff
@@ -313,7 +315,162 @@ Notes:
...
+				noQualifierDiscounts.promotionID
+			FROM
+				noQualifierDiscounts, orderItemDataQuery 
+			WHERE
+				orderItemDataQuery.skuID = noQualifierDiscounts.skuID
+		</cfquery>
+		
+		<cfloop query="orderItemDiscountsQuery">
+			<cfif orderItemDiscountsQuery.salePriceDiscountType EQ "amount">
+				<cfset orderItemDiscountsQuery.salePrice = orderItemDiscountsQuery.amount >
```
sumitverma: I don't think you set query columns like that. You needs to use querySetCell()

christopher-kundrat: Yeah I wasn't sure myself so I tested it to check and the value was getting set. I can update it if you want though.

christopher-kundrat:
Closing so that this can be merged into hotfix.

PR #3130, opened 2015-03-31
Account email address verification

UltimoTG:
Send out verification email only if the email address has not been verified


gregmoser:
Ryan, this is good to merge.

PR #3171, opened 2015-04-07
Fix for directory paths used to copy Slatwall modify templates for Mura ...

chriskent:
...install

### DIFF
--- a/integrationServices/mura/model/handler/MuraEventHandler.cfc
+++ b/integrationServices/mura/model/handler/MuraEventHandler.cfc
@@ -769,10 +769,11 @@
 				if(getMuraPluginConfig().getSetting("createDefaultPages") && !listFindNoCase(populatedSiteIDs, cmsSiteID)) {
 					
 					// Copy views over to the template directory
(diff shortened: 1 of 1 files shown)


integrationServices/mura/model/handler/MuraEventHandler.cfc
```diff
@@ -769,10 +769,11 @@
 				if(getMuraPluginConfig().getSetting("createDefaultPages") && !listFindNoCase(populatedSiteIDs, cmsSiteID)) {
 					
 					// Copy views over to the template directory
-					var slatwallTemplatePath = getDirectoryFromPath(expandPath("/Slatwall/public/views/templates/")); 
-					var muraTemplatesPath = getDirectoryFromPath(expandPath("/muraWRM/#cmsSiteID#/includes/themes/#cmsThemeName#/templates/"));
+					var slatwallTemplatePath = getDirectoryFromPath(expandPath("/Slatwall/public/views/templates"));
```
sumitverma: Might be better to change these to:
getDirectoryFromPath(expandPath("/Slatwall") & "/public/views/templates")

and same for /muraWRM. 

Expand path just seem to work dirrefent on different environment, but always works consistently when we use a know root mapping/path.

PR #3188, opened 2015-04-14
Added ability to added custom attributes to option groups

DavidShaughnessy:


sumitverma:
You also need to add the properties and bidirectional methods. Here is a commit you can use as a reference:

https://github.com/ten24/slatwall/commit/65a48feaaa5bd7364709ff105b4a965a1b569b75

PR #3249, opened 2015-05-06
Just needed to run the array sorter utility against name before type

DavidShaughnessy:


admin/client/js/es5/directives/productBundleGroup/swproductbundlegroup.js
```diff
@@ -242,6 +242,7 @@ angular.module("slatwalladmin").directive("swProductBundleGroup", ["$http", "$lo
 				//Pushes item back into array
 				scope.productBundleGroupFilters.value.push(scope.productBundleGroup.data.skuCollectionConfig.filterGroups[0].filterGroup[index]);
 				//Sorts Array
+				scope.productBundleGroupFilters.value = utilityService.arraySorter(scope.productBundleGroupFilters.value, "name");
```
rmarchandTen24: i believe this won't do anything because the line below this one will override the sorting

DavidShaughnessy:
It first sorts all of the records by name alphabetically, then groups the alphabetized records by type.

PR #3257, opened 2015-05-12
Hotfix ds arraysorter

DavidShaughnessy:


rmarchandTen24:
instead of array of array. use javascript object
{
 'key':"string",
"direction":'asc' |'desc'
}

PR #3301, opened 2015-05-29
simplifying code to maintain the "Material Paths" aka URLTitlePath for

rmarchandTen24:
children

### DIFF
--- a/model/entity/Content.cfc
+++ b/model/entity/Content.cfc
@@ -159,30 +159,55 @@ component displayname="Content" entityname="SlatwallContent" table="SwContent" p
 	}
 	
 	public string function createURLTitlePath(){
+		
 		var urlTitle = '';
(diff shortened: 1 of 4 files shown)


model/entity/Content.cfc
```diff
@@ -159,30 +159,55 @@ component displayname="Content" entityname="SlatwallContent" table="SwContent" p
...
+	
 	public string function setUrlTitle(required string urlTitle){
+		//look up all children via lineage
+		var previousURLTitlePath = '';
+		if(!isNull(this.getURLTitlePath())){
+			previousURLTitlePath = this.getURLTitlePath();
+		}
+		 
+		var allDescendants = getAllDescendants();
+		//set url title
```
sumitverma: getAllDescendants could potentially return 100's of object, I would recommend using query here instead of object to update. We could potentially do this in one update query.

PR #3370, opened 2015-06-23
3 part fix for update calculated properties task

chriskent:
1. thread request account does not have a primary address set
2. when run as scheduled task task history not being saved after run, so never setting end time/success flag
3. when run as scheduled task & task schedule end set to "forever" then endDateTime was undefined so final stage of scheduled task was aborting on error with no log/error email sent

### DIFF
--- a/model/entity/Schedule.cfc
+++ b/model/entity/Schedule.cfc
@@ -112,7 +112,7 @@ component displayname="Schedule" entityname="SlatwallSchedule" table="SwSchedule
 	public string function getNextRunDateTime(startDateTime, endDateTime){
 		var nextRun='';
 		
-		if( !isDefined('arguments.endDateTime') or endDateTime > now() ){
+		if( isNull(arguments.endDateTime) or endDateTime > now() ){
 			
 			switch(getRecuringType()){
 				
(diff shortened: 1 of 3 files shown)


model/entity/Schedule.cfc
```diff
@@ -112,7 +112,7 @@ component displayname="Schedule" entityname="SlatwallSchedule" table="SwSchedule
 	public string function getNextRunDateTime(startDateTime, endDateTime){
 		var nextRun='';
 		
-		if( endDateTime > now() ){
+		if( !isDefined('arguments.endDateTime') or endDateTime > now() ){
```
rmarchandTen24: isNull(arguments.endDateTime)

model/service/TaskService.cfc
```diff
@@ -174,6 +174,7 @@ component extends="HibachiService" output="false" accessors="true"{
 			taskHistory.setEndTime( now() );
 			
 			// Persist the info to the DB
+			getHibachiDAO().save(taskHistory);
```
rmarchandTen24: replace with this.saveTaskHistory( taskHistory )

PR #3383, opened 2015-06-26
Display Sku Count in View Product screen.

rylan-doherty:
Added information to the count attribute so that it displays Sku Count.


gregmoser:
This should be a pull request to 'hotfix' and not 'master' so that the CI environment can run.

PR #3385, opened 2015-06-26
adding fix to skucode

rylan-doherty:


rmarchandTen24:
update to use the loop index

PR #3386, opened 2015-06-26
Hotfix rm updating dao method for count

rmarchandTen24:
updating the dao method for get#property#count to use smartlist recordsCount

### DIFF
--- a/org/Hibachi/HibachiEntity.cfc
+++ b/org/Hibachi/HibachiEntity.cfc
@@ -527,7 +527,7 @@ component output="false" accessors="true" persistent="false" extends="HibachiTra
 		var cacheKey = "#arguments.propertyName#Count";
 			
 		if(!structKeyExists(variables, cacheKey)) {
-			variables[ cacheKey ] = arrayLen(variables[ arguments.propertyName ]);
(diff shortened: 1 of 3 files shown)


org/Hibachi/HibachiEntity.cfc
```diff
@@ -524,13 +524,8 @@ component output="false" accessors="true" persistent="false" extends="HibachiTra
...
 	// @hint returns the count of a given property
 	public numeric function getPropertyCount( required string propertyName ) {
-		var cacheKey = "#arguments.propertyName#Count";
-			
-		if(!structKeyExists(variables, cacheKey)) {
-			variables[ cacheKey ] = arrayLen(variables[ arguments.propertyName ]);
-		}
-		
-		return variables[ cacheKey ];
+		var propertySmartList = this.invokeMethod('get#propertyName#SmartList');
```
sumitverma: I would recommend using the old code, because that will be more performant as arraylen will do count(*) anyways. Smartlist will have more overhead and if the smartlist is overriden you can even get unexpected result. 

We can certainly create another method that gets count based on smartlist.

PR #3388, opened 2015-06-29
Add optionGroup.sordOrder property to getOptionsSmartlist 

chriskent:
to fix MS SQL "ORDER BY items must appear in the select list if SELECT DISTINCT is specified"


sumitverma:
When add select is added, it will _only_ return a hash map of the properties in select list. So, I don't think will fix anything, this should actually make this method unusable as it will only return sortorder now.

chriskent:
Remove property sortOrder from optionGroup orderBy  to fix MS SQL "ORDER BY items must appear in the select list if SELECT DISTINCT is specified"

This works and keeps options grouped by option group but does not force the orderGroup sort order

PR #3391, opened 2015-06-30
validating to make sure attributes do not have hyphens

rmarchandTen24:


config/resourceBundles/en.properties
```diff
@@ -210,11 +199,11 @@ admin.entity.integrationtabs.settings=Integration Settings
 admin.entity.listAudit=Audit Log
 admin.entity.listLoyalty=Loyalty Program
 admin.entity.listapp=Applications
-admin.entity.listcontent=Contents
 admin.entity.listcartandquote=Carts & Quotes
 admin.entity.listcartandquote_nav=Carts & Quotes
 admin.entity.listcartandquote_permission=Carts & Quote - View List
 admin.entity.listcartandquote_title=View A List of Carts & Quotes
+admin.entity.listcontent=Contents
```
sumitverma: This should be Content

gregmoser:
Why can't attribute codes have hyphens?

chriskent:
One reason I have found is that you cannot use shorthand to get attributes if the code has hyphens
getMyEntity().getMy-Attribute-Code()
but you can use
getMyEntity().getMyAttributeCode()

gregmoser:
I just spoke to Ryan & Sumit who informed me that this issue is with HQL... because it can't do aliases with hyphens, but this is another good point.

PR #3410, opened 2015-07-01
Fix for sku.getOptionsByOptionGroupIDStruct()

dzikakis:
Looks like there was a typo in here


rmarchandTen24:
Thanks again Dave,

I issued a new pull request just so it is readable by github
https://github.com/ten24/slatwall/pull/3435

PR #3418, opened 2015-07-02
Fix for asset paths - so they show up in a mura install.

dzikakis:


rmarchandTen24:
Hi Dave, is it possible to fix this with a Slatwall Mapping instead?

dzikakis:
They're all static assets, so I suppose it would be possible to add mappings in the web server (apache/iis/nginx or whatever), but that seems like a major pain. Why can't we use the relative paths in the css? The change in login.cfm is consistent with other paths in slatwall too. Seems like this would be the best option for portability IMO.

rmarchandTen24:
fix was applied

PR #3425, opened 2015-07-02
Modified regex validation for attribute code with corresponding language

jasperboyd:
Modified regex for attribute code to deny codes beginning with a number, also added the corresponding language to rb. Fixed a typo while I was at it.


rmarchandTen24:
use this ^[a-zA-Z][a-zA-Z0-9_]*$

to exclude weird characters on the first character

update the rbkey to explain why

PR #3428, opened 2015-07-07
updating path to javascript when not in debug

jasperboyd:


admin/layouts/default.cfm
```diff
@@ -375,13 +375,14 @@ Notes:
 
 		<script type="text/javascript" src="#request.slatwallScope.getBaseUrl()#/?slatAction=api:js.ngslatwall&instantiationKey=#$.slatwall.getApplicationValue('instantiationKey')#"></script>
 		<cfif request.slatwallScope.getApplicationValue('debugFlag')>
-			<cfset es5scriptPath = expandPath('/admin/client/js/es5/')>
+			<cfset es5scriptPath = expandPath('/#request.slatwallScope.getDao('hibachiDao').getApplicationKey()#/admin/client/js/es5/')>
```
sumitverma: It's better to just do expandPath on root folder and append. this avoids issue with expan path when folder doesn't exists or permission issues. So:

<cfset es5scriptPath = expandPath('/#request.slatwallScope.getDao('hibachiDao').getApplicationKey()#/') & "admin/client/js/es5/">

Issue #3432, opened 2015-07-07
Account without email address causes error when placing order


jsieber-aca:
Slatwall Version:	3.5.061
CFML Server:		ColdFusion Server: 10,0,16,293499
DB Dialect: 		MySQL

When an account is created within the admin, it does not require an email address and this causes errors when an order is placed under that account.

When using the admin for walkin and phone orders, not all customers will provide an email address, so we prefer to keep it optional, but it does throw an error when trying to place an order under the account without an email. The error occurs in the emailService.cfc as the cfmail tag requires that a to address is defined, but is passed as an empty string if the account email was not set when the account is created.

Steps to reproduce:
1. Create a new account in admin without defining an email address.
2. Place a new order under the account created in step 1.
3. Error will occur in sendEmail function in EmailService.cfc as arguments.email.getEmailTo() returns and empty string, which is invalid as the to attribute for cfmail.

jsieber-aca:
Thanks!

PR #3433, opened 2015-07-07
Resolves issue with orders placed under an account without an email a…

jsieber-aca:
…ddress

Fixes issue https://github.com/ten24/slatwall/issues/3432


rmarchandTen24:
thanks John

PR #3434, opened 2015-07-08
Patch 25

dzikakis:
This is causing an issue "can't cast Null value to value of type [struct]". setting the data argument to type ANY fixes it so it can accept nulls. Noticed this when trying to add something to the cart from the front end.

### DIFF
--- a/org/Hibachi/Hibachi.cfc
+++ b/org/Hibachi/Hibachi.cfc
@@ -701,7 +701,7 @@ component extends="FW1.framework" {
 	}
 	
 	// This method will execute an actions controller, render the view for that action and return it without going through an entire lifecycle
-	public string function doAction(required string action, struct data) {
+	public string function doAction(required string action, any data) {
 		
 		var response = "";
 		var originalFW1 = {};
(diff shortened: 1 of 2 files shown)


rmarchandTen24:
Hi Dave,
can you tell me the action you called when you received the error?

rmarchandTen24:
i know you said add to cart but I wanted to know was it public:cart.addOrderItem?

dzikakis:
Pretty sure it was public:cart.addOrderItem as that's all I think I use on the front, it was last week that I ran into this, I can double check again shortly.

rmarchandTen24:
I think that maybe line 121 of cart.cfc should have been declared a struct
and maybe that is the cause of the issue?

https://github.com/ten24/slatwall/blob/master/public/controllers/cart.cfc#L121

rmarchandTen24:
(required any rc) should be (required struct rc) me thinks

dzikakis:
Taking another look at this.. I think it might be the MuraEventHandler that needs updating. ?

dzikakis:
Could also just change the argument to: struct data={}, so the default is an empty struct. I have no idea if anyone would have created an integration that calls doAction() ?

PR #3447, opened 2015-07-10
client to server side logging with angular

jasperboyd:


org/Hibachi/HibachiControllerREST.cfc
```diff
@@ -537,6 +539,29 @@ component output="false" accessors="true" extends="HibachiController" {
...
+		}
+		var cause = 'There was no reported cause';
+		if(structKeyExists(arguments.rc.deserializedJSONData,'cause')){
+			cause = arguments.rc.deserializedJsonData.cause;
+		}
+		try { 
+			throw(type="ClientError", message="Exception: #exception# Cause: #cause#");
+		} catch (any e) { 
+			//pass through
+		}
```
sumitverma: You don't need try catch around throw...

PR #3458, opened 2015-07-15
Feature mt aggregate collections

migueltarga:
Created the Aggregate helper methods


admin/client/ts/directives/content/swcontentlist.ts
```diff
@@ -72,6 +72,13 @@ angular.module('slatwalladmin')
                             ormtype:'boolean',
                             isSearchable:false
                         }
+//                        ,{
```
rmarchandTen24: can you remove the comments on this. I thiink it was an unrelated change

meta/tests/unit/entity/CollectionTest.cfc
```diff
@@ -104,7 +177,7 @@ component extends="Slatwall.meta.tests.unit.entity.SlatwallEntityTestBase" {
 		};
 		
 		var aggregateHQL = variables.entity.getAggregateHQL(aggregate,propertyIdentifier);
-		//request.debug(aggregateHQL);
+		request.debug(aggregateHQL);
```
rmarchandTen24: we should remove the commented code and all request.debugs before pushing up the test

model/entity/Collection.cfc
```diff
@@ -746,7 +837,7 @@ component displayname="Collection" entityname="SlatwallCollection" table="SwColl
 	
 	// Paging Methods
 	public array function getPageRecords(boolean refresh=false) {
-		try{
+		//try{
```
rmarchandTen24: we need this try catch for UI purposes so this should be added back in

PR #3464, opened 2015-07-16
Just using conditionals to hide the buttons. Updating the validation

christopher-kundrat:
caused problems.


gregmoser:
This is no good... because it leaves things super open for issues in the future.  Also, there are 2 places that these buttons show up.

PR #3520, opened 2015-08-07
Feature subscription usages updates

ian-a-hickey:


model/entity/Account.cfc
```diff
@@ -696,4 +695,27 @@ component displayname="Account" entityname="SlatwallAccount" table="SwAccount" p
...
+	
+	
+	/*
+		test
+	*/
+		
+	

+	//CUSTOM FUNCTIONS END 
 }
```
sumitverma: Please remove the test methods.

model/entity/PaymentMethod.cfc
```diff
@@ -213,6 +213,20 @@ component entityname="SlatwallPaymentMethod" table="SwPaymentMethod" persistent=
...
+			if (subscriptionTerm.getAutoPayFlag()){
+				foundAutoPayFlagSet = true;
+				break;
+			}
+		}
+		if (foundAutoPayFlagSet){
+			return false;
+		}
+		return true;
+	}
```
sumitverma: Instead of looking through all terms, you can just add a filter and do smartlist.getRecordsCount()

model/entity/SubscriptionTerm.cfc
```diff
@@ -91,14 +91,32 @@ component entityname="SlatwallSubscriptionTerm" table="SwSubscriptionTerm" persi
...
+		for (var paymentMethod in activePaymentMethods){
+			if (paymentMethod.getAllowSaveFlag()){
+				foundPaymentMethodThatAllowsAccountsToSave = true;
+				break;
+			}
+		}
+		return foundPaymentMethodThatAllowsAccountsToSave;
+	}
+	
 }
```
sumitverma: Instead of looping through all methods, you can just add a filter and do smartlist.getRecordsCount()

model/entity/SubscriptionUsage.cfc
```diff
@@ -158,14 +170,74 @@ component entityname="SlatwallSubscriptionUsage" table="SwSubsUsage" persistent=
...
+	public any function hasSubscriptionOrderItems(){
+		if ( arrayLen( getSubscriptionOrderItems( ) ) ) {
+			return true;
+		}
+		return false;
+	}
+	
+	public any function getInitialSubscriptionOrderItem(){
+		if( hasSubscriptionOrderItems() ){
+			return getSubscriptionOrderItems()[1];
```
sumitverma: I don't think getting the first orderitem will work because the records are not always sorted. You will need to do a smartlist with order item type as initial.

model/entity/SubscriptionUsage.cfc
```diff
@@ -158,14 +170,74 @@ component entityname="SlatwallSubscriptionUsage" table="SwSubsUsage" persistent=
...
+		}
+	}
+	
+	public any function getInitialOrderItem(){
+		if( hasSubscriptionOrderItems() && !isNull( getSubscriptionOrderItems()[1].getOrderItem() )){
+			getSubscriptionOrderItems()[1].getOrderItem();
```
sumitverma: same as above, but just delegate to getInitialSubscriptionOrderItem().getOrderItem() for all the initialxxx methods.

model/entity/SubscriptionUsage.cfc
```diff
@@ -158,14 +170,74 @@ component entityname="SlatwallSubscriptionUsage" table="SwSubsUsage" persistent=
...
+	public any function getInitialOrder(){
+		if( hasSubscriptionOrderItems() && !isNull( getSubscriptionOrderItems()[1].getOrderItem().getOrder() )){
+			return getSubscriptionOrderItems()[1].getOrderItem().getOrder();
+		}
+		
+	}
+	
+	public any function getMostRecentSubscriptionOrderItem(){
+		if( hasSubscriptionOrderItems() && !isNull( getSubscriptionOrderItems()[ getTotalNumberOfSubscriptionOrderItems() ] )){
+			return getSubscriptionOrderItems()[ getTotalNumberOfSubscriptionOrderItems() ];
```
sumitverma: use smartlist and order by date

model/service/OrderService.cfc
```diff
@@ -75,6 +75,23 @@ component extends="HibachiService" persistent="false" accessors="true" output="f
...
+				for (var orderPayment in arguments.order.getOrderPayments()){
+					if (!isNull(orderPayment.getAccountPaymentMethod())){
+						hasOrderPaymentWithAccountPaymentMethod = true;
+					}
+				}
+			}
+		}
+		if (hasSubscriptionWithAutoPay && !hasOrderPaymentWithAccountPaymentMethod){
+			orderRequirementsList = listAppend(orderRequirementsList, "accountPaymentMethod");
+		}
```
sumitverma: I don't think we need this. AccountPaymentMethod is not required to place the order. It only needs to be set when the order is placed and it's an autopay order.

model/service/OrderService.cfc
```diff
@@ -1112,6 +1149,9 @@ component extends="HibachiService" persistent="false" accessors="true" output="f
 					// Verify the order requirements list, to make sure that this order has everything it needs to continue
 					if(len(orderRequirementsList)) {
 						
+						if(listFindNoCase("accountPaymentMethod", orderRequirementsList)) {
+							arguments.order.addError('payment', rbKey('entity.order.process.placeOrder.hasSubscriptionWithAutoPayFlagWithoutOrderPaymentWithAccountPaymentMethod'));
+						}
```
sumitverma: new order requirement not needed.

model/service/OrderService.cfc
ian-a-hickey: I thought we needed this because the creation of the subscription usage will depend on it.

sumitverma: no, the subscription usage will get created irrespective of payment methods. Just setting setSaveAccountPaymentMethodFlag (which you are already doing) will take care of this requirement.

PR #3538, opened 2015-08-12
added a dblclickevent which surpresses form resubmission

jasperboyd:
does not disable the button on invalid input


rmarchandTen24:
this only solves quick bursts. It doesn't solve a click wait and click again which occurs when a long process makes someone impatient.

PR #3539, opened 2015-08-12
changing redirect to main page

jasperboyd:
per gregs suggestion


rmarchandTen24:
this will redirect after all form submits regardless of session expired

PR #3549, opened 2015-08-14
Hotfix jb order payment amount js

jasperboyd:
replicated old functionality


gregmoser:
In case this fails for whatever reason, just go back to this version of the logic and remove the default 'hide' classes from the dom:

https://github.com/ten24/slatwall/blob/5744aa349e114e7b66749285721d90f45b342d0a/admin/views/entity/preprocessorder_addorderpayment.cfm

PR #3605, opened 2015-09-01
Removes 'invalid' or 'removed' orderPayments for return order options

gregmoser:


gregmoser:
This has not been tested, it is just here for reference.

PR #3611, opened 2015-09-03
Hotfix jb authorizenet endpoint settings

jasperboyd:


rmarchandTen24:
why isn't the fieldtype specified?

rmarchandTen24:
explicitly set this to text please

PR #3612, opened 2015-09-03
added validation to prevent account dependent order actions when the

jasperboyd:
account is unassigned


gregmoser:
Account is NOT required when adding and orderItem.  That would be like 'addToCart' on the frontend.

sumitverma:
Right, Account should be required for add order payment but not for order item.

jasperboyd:
I'll change and resubmit, I'm going to open up a new story then to fix the order item view because it throws an error when no account is attached as well

PR #3616, opened 2015-09-08
download file redirect

jasperboyd:


rmarchandTen24:
this code will not work because getFileService

PR #3625, opened 2015-09-09
auth override on collections for developers

jasperboyd:


gregmoser:
I'm a bit confused by this.  I don't know why you would need it at all.  Unit tests should run as a super user unless the it's a specific test that is intentionally testing permissions.

PR #3634, opened 2015-09-14
giftcard detail page wired with angular

jasperboyd:
only missing actions pending review


admin/client/partials/entity/GiftCard/basic.html
```diff
@@ -0,0 +1,23 @@
+<div class="col-md-12">
+	<div class="alert alert-danger alert-dismissible" role="alert">
+		<button type="button" class="close" data-dismiss="alert" aria-label="Close"><span aria-hidden="true">&times;</span></button>
+		<strong>Recipient Email Failure!</strong> There was an issue sending your gift card to the provided recipient email <button class="btn btn-danger btn-xs">Resend Recipient Email</button>
+	</div>
+	<div class="row s-detail-modules-wrapper">
```
rmarchandTen24: why don't we just use one ng-if="giftCard" on the wrapping div?

admin/client/partials/entity/GiftCard/history.html
```diff
@@ -0,0 +1,158 @@
...
+								<th>Type <i class="fa fa-sort"></i></th>
+								<th>Date <i class="fa fa-sort"></i></th>
+								<th>Info</th>
+								<th>Transaction</th>
+								<th>Balance</th>
+							</tr>
+						</thead>
+						<tbody>
+							<tr ng-class="{'s-purchase':transaction.debit, 's-success':transaction.activated}" ng-repeat="transaction in transactions">
+								<!--- Initial Transactions --->
```
rmarchandTen24: not a big issue but i think that github is formatting this incorrectly because you are using coldfusion commenting style in a html file.  is used for html comments instead of

admin/client/partials/entity/GiftCard/history.html
```diff
@@ -0,0 +1,158 @@
...
+								<td ng-if="$index==transactions.length-1">Customer: <a href="##"><span ng-bind="order.account_firstName"></span> <span ng-bind="order.account_lastName"></span> (<span ng-bind="order.account_primaryEmailAddress_emailAddress"></span>)</a></td>
+								<td ng-if="$index==transactions.length-1">
+									+ <span ng-bind="transaction.creditAmount"></span>
+								</td>
+								<td ng-if="$index==transactions.length-1">
+									<span ng-bind="transaction.creditAmount"></span>
+								</td>
+								
+								<!--- Notification Email --->
+								<td ng-if="transaction.emailSent"><i class="fa fa-envelope"></i></td>
```
rmarchandTen24: in a case like this where no obvious wrapper would solve the ng-if sibling issue. I recommend using ng-if-start and ng-if-end. This only sets up one watcher. example:
<td ng-if-start="transaction.emailSent" >Recipient notification sent </td>
+								<td ng-bind="transaction.sentAt"></td>
+								<td>Recipient: <span ng-bind="giftCard.orderItemGiftRecipient_firstName"></span> <span ng-bind="giftCard.orderItemGiftRecipient_lastName"></span> ( <span ng-bind="giftCard.orderItemGiftRecipient_emailAddress"></span> )</td>
+								<td ng-if-end>--</td>

The -start and -end suffix can be applied to most directives and can be set up for custom directives as well

admin/client/ts/directives/giftcard/swgiftcardbalance.ts
```diff
@@ -0,0 +1,60 @@
+module slatwalladmin { 
+	'use strict'; 
+	
+	export class GiftCardBalance implements ng.IDirective { 
+		
+		public static $inject = ["$slatwall", "$templateCache", "partialsPath"];
+		public restrict:string; 
+		public templateUrl:string;
+		public scope = {
```
rmarchandTen24: scope should remain {} and bind to controller should contain all scope items. see swpagination for details. This is going to be necessary for the angular 2.0 migration

admin/client/ts/directives/giftcard/swgiftcardbalance.ts
```diff
@@ -0,0 +1,60 @@
...
+		public templateUrl:string;
+		public scope = { 
+			giftCard:"=?", 
+			transactions:"=?", 
+			initialBalanceFormatted:"=?", 
+			currentBalanceFormatted:"=?",
+			balancePercentage:"=?"
+		}; 
+		public bindToController; 
+
```
rmarchandTen24: add public controller:<insert controller class>
and public controllerAs:<insert namespace ex: SwGiftCardBalance>

you will then use SwGIftCardBalance.<scope variable> in the partial to access the items. This will ensure there are no conflicted scopes. Adding a dot to scopes is a best practice and ensures binds update correctly

PR #3635, opened 2015-09-14
Fix Collection keywords search when columns is not specified

migueltarga:


model/entity/Collection.cfc
```diff
@@ -1258,338 +1258,255 @@ component displayname="Collection" entityname="SlatwallCollection" table="SwColl
...
+					defaultColumns && (
+						structKeyExists(column, 'fkcolumn')
+					|| (structKeyExists(column, 'persistent') && column.persistent == false)
+					|| !structKeyExists(column, 'ormtype')
+					))
+				) continue;
+				//if ormtype is not set, find it
+				if(!structKeyExists(column, 'ormtype')){
+					var allColumns = getService('HibachiService').getPropertiesWithAttributesByEntityName(arguments.collectionConfig.baseEntityName);
+					for( col in allColumns){
```
rmarchandTen24: need to var scope the col variable

PR #3647, opened 2015-09-21
retrieve entity name in a request

migueltarga:
Retrieve correct entityName from request when it doesn't follow the conventions


sumitverma:
This will not work when multiple IDs are passed or if the passed ID is not related to action. e.g. accountID=xx&addressID=yy

migueltarga:
Sumit is right, I will just close this pull request until I find a better solution.

PR #3650, opened 2015-09-22
fixes how update calculated properties works so that when multiple

ian-a-hickey:
transactions happen during a single request, there is a way for allowing
the updated to fire multiple times as needed.

The issue: If more than one transaction was firing during a request such as removing an orderitem before adding an orderitem, if an update calculated properties was called more than once, it was only firing one time. This pull request allow a person to explicitly update calculated properties more than once if needed.
You can do that, by setting setUpdateRunFlag(false) before calling updateCalculatedProperties().

Searched for references for that variable that had its name changed: CalculatedUpdateRunFlag and didn't find any.

### DIFF
--- a/model/service/OrderService.cfc
+++ b/model/service/OrderService.cfc
@@ -1495,6 +1495,11 @@ component extends="HibachiService" persistent="false" accessors="true" output="f
 			
 			// Re-Calculate tax now that the new promotions and price groups have been applied
 			getTaxService().updateOrderAmountsWithTaxes( arguments.order );
+			
+			//update the calculated properties
+			arguments.order.setUpdateRunFlag(false);
+			arguments.order.updateCalculatedProperties();
+			
 		}
 		return arguments.order;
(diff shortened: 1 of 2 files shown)


gregmoser:
What if we put this in the hibachiScope, and did something like this:

if(!getHibachiScope().getCalculationUpdateRun( this.getPrimaryID() ))
getHibachiScope().addCalculationUpdateRun( this.getPrimaryID() )
}

And then the hibachiScope just keeps an array of ID's that calculations were run on.  At the end of every action in a chained action list we just clear out the array of ID's in the hibachiScope.  Basically this would be part of the doAction() function in Hibachi.cfc

Then this would just work everywhere without having to explicitly call it out.

Thoughts?

ian-a-hickey:
This makes a bunch of sense. I'll make that change.

rmarchandTen24:
closed pending revisions

PR #3659, opened 2015-09-25
Update Gift Card Email When There is A Email Bounce

jasperboyd:


admin/client/ts/directives/giftcard/swgiftcardhistory.ts
```diff
@@ -81,11 +78,13 @@ module slatwalladmin {
...
 						
-						angular.forEach(this.bouncedEmails, (email, bouncedEmailIndex)=>{
-							email.bouncedEmail = true; 
-							email.balanceFormatted =  "$" + parseFloat(initialBalance.toString()).toFixed(2);
-							this.transactions.splice(index, 0, email);
-						}); 
+						if(this.bouncedEmails.length > 0){
+							angular.forEach(this.bouncedEmails, (email, bouncedEmailIndex)=>{
+								email.bouncedEmail = true; 
+								email.balanceFormatted =  "$" + parseFloat(initialBalance.toString()).toFixed(2);
```
rmarchandTen24: you should look into using a currency filter on the template instead of concatenating the string

https://docs.angularjs.org/api/ng/filter/currency

model/service/GiftCardService.cfc
```diff
@@ -107,10 +107,10 @@ component extends="HibachiService" persistent="false" accessors="true" output="f
 		}
 
 		if(!giftCardCreditTransaction.hasErrors()){
```
rmarchandTen24: When this.saveGiftCard runs it will automatically run .validate so this is probably redundant. 
it should be
arguments.giftCard = this.saveGiftCard(arguments.giftCard);
if(arguments.giftCard.hasErrors()){
...
}

admin/client/ts/directives/giftcard/swgiftcardhistory.ts
gregmoser: I'm not sure how ng supports multiple currency formatting, but formatted currency might need to come from the server side because we have a lot of logic based on business rules to determine how currency should output.

Please always keep in mind that a lot of users are NOT using USD.

PR #3663, opened 2015-09-25
fix for railo and luceee in email bounce service

jasperboyd:


gregmoser:
Lets pick a better place for these files than just the root /model directory.  Also, I'd like to test and see if defining the properties in an attributeCollection definition solves the problem:

```
<cfset imapAttributes = structNew() />
....
<cfimap attributeCollection=imapAttributes />
```

PR #3668, opened 2015-09-28
Update EmailBounceService

ian-a-hickey:
Added a ; to line 83 to fix hard error in CF


ian-a-hickey:
Pulls down feature. hard error. this fixes it. editing from github directly but still submitting pull request.

PR #3673, opened 2015-09-29
Feature ck add remote id to vendor entities

christopher-kundrat:


brucebailey:
Due to an overwhelming number of junk messages, this email address is walking among the dead.

If you are a real person with a real message for me, please resend to bailey at movementdh dot com.  

Thank you!

PR #3677, opened 2015-09-30
now displaying gift cards in order delivery detail

jasperboyd:


rmarchandTen24:
can we go over this smartlist? We may be able to make this simpler

model/entity/OrderDelivery.cfc
```diff
@@ -84,6 +84,18 @@ component displayname="Order Delivery" entityname="SlatwallOrderDelivery" table=
 		return totalDelivered;
 	}
 
+	public any function getOrderDeliveryGiftCardSmartList() {
+		var giftCardSmartList = getService("HibachiService").getGiftCardSmartList();
```
gregmoser: You should be getting this NOT for getService("hibachiService") but it should be getService("GiftCardService")

jasperboyd: Doesn't the get#entity#smartlist() come from hibachi service? or is this more of a convention?

rmarchandTen24: yes true but going through the giftcardservice is in preparation for an override

PR #3681, opened 2015-10-02
wiring up gc detail links

jasperboyd:


rmarchandTen24:
there are <<<<<<< HEAD blocks in the code. can you remove those and verify the conflict was resolved correctly

PR #3707, opened 2015-10-15
unit tests for creating a gc product with new test feature add entity to tear down

jasperboyd:


gregmoser:
Case 1, 2 & 3... should be 3 different unit test functions.  You shouldn't loop different cases into the same unit test.  If you need a private function that all 3 use for setup that is fine.

This is all IMO (Ryan it's on you to decide if you agree or not)

rmarchandTen24:
exactly, you can prefix the function and underscore the case
createGiftCardProduct_percentage.
then a private function for setup createGiftCardProduct_setup, which will
be called as a private funciton by the other cases

PR #3711, opened 2015-10-20
opening up productSmartlist to search by skucode

rmarchandTen24:


gregmoser:
This has performance implications and needs to be thought out.

rmarchandTen24:
right I mentioned that as well and it is not going into core.

PR #3736, opened 2015-10-28
Feature ck option group by product type error

christopher-kundrat:


rmarchandTen24:
use cfqueryparam of bit

PR #3753, opened 2015-10-30
Client Side Validation For Add Recipients, GC Detail Bug Fixes, Order Cancellation GC Refund

jasperboyd:


model/process/GiftCard_Create.cfc
```diff
@@ -71,6 +71,18 @@ component output="false" accessors="true" extends="HibachiProcess"{
 		}
 	}
 
+
+	public string function getGiftCardPin(){
+		if(isNull(variables.giftCardPin)){
+			var pin = "";
+			for(var i=1; i<4; i++){
+				pin = pin & RandRange(0, 9);
+			}
```
sumitverma: Is there a benefit to loop than just RandRang(100,999)?

jasperboyd: @sumitverma I don't think there's any difference I can change it to that

PR #3757, opened 2015-11-03
adding update script for auto fulfill flag

jasperboyd:


config/scripts/database/v4_2.cfm
```diff
@@ -0,0 +1,77 @@
...
+<!--- Update SwContent create materialized paths for title called titlePath based on existing title --->
+<cftry>
+
+	<cfquery name="local.updateautofulfilltrue" dbtype="query">
+		update SwFulfillmentMethod set autoFulfillFlag=1 where fulfillmentMethodType=<cfqueryparam cfsqltype="cf_sql_varchar" value="email"> or fulfillmentMethodType=<cfqueryparam cfsqltype="cf_sql_varchar" value="auto">
+	</cfquery>
+
+	<cfquery name="local.updateautofulfillfalse" dbtype="query">
+		update SwFulfillmentMethod set autoFulfillFlag=0 where fulfillmentMethodType!=<cfqueryparam cfsqltype="cf_sql_varchar" value="email"> and fulfillmentMethodType!=<cfqueryparam cfsqltype="cf_sql_varchar" value="auto">
+	</cfquery>
```
sumitverma: I think we should set these value only if it's null. That way we don't override it once user sets it.

PR #3760, opened 2015-11-04
Creating Email Fulfillment From Sku Template & Other Email Improvements

jasperboyd:


rmarchandTen24:
uses org/hibachiservicet getPrimaryIDPropertyNameByEntityName
and chang getsku to use newflag or null checking

PR #3777, opened 2015-11-11
order item validation and associated rb key

jasperboyd:


rmarchandTen24:
needs conflict resolution

PR #3778, opened 2015-11-12
listing display full text search

jasperboyd:


admin/client/ts/services/collectionconfigservice.ts
```diff
@@ -248,13 +248,43 @@ module slatwalladmin{
                 }
                 if(!angular.isUndefined(options['ormtype'])){
                     ormtype = options['ormtype'];
-                }else if(this.collection.metaData[lastProperty] && this.collection.metaData[lastProperty].ormtype){
+                }else if(properties.length === 2 && this.collection.metaData[lastProperty] && this.collection.metaData[lastProperty].ormtype){
```
rmarchandTen24: jasper, just a quick question? What are we using properties.length === 2 for?

jasperboyd: if there are no relationships the property string is going to be entity.property > 2 it means theres a relationship and we need to traverse all the metadata

rmarchandTen24: can you supply a concrete case?

jasperboyd: _product.defaultSku.price

PR #3826, opened 2015-11-30
Renewal Price Changes 

jasperboyd:


admin/client/ts/controllers/preprocessproduct_create.ts
```diff
@@ -5,61 +5,83 @@ module slatwalladmin {
...
-                })
-            }
-            
-            var jQueryOptionsRedemptionAmountType = $("select[name='redemptionAmountType'")[0];
-            this.$scope.preprocessproduct_createCtrl.redemptionAmountTypeOptions = [];
-            angular.forEach(jQueryOptionsRedemptionAmountType,(jQueryOption)=>{
-                var option = {
-                    label:jQueryOption.label,
-                    value:jQueryOption.value    
+<<<<<<< HEAD
```
rmarchandTen24: jasper can you remove the conflict tag

jasperboyd: woops good catch it's outta there now

PR #3833, opened 2015-11-30
taking out first and last name from gift recipient

jasperboyd:


admin/client/partials/entity/OrderItemGiftRecipient/addorderitemgiftrecipient.html
```diff
@@ -9,11 +9,11 @@
 		<table class="table table-bordered table-hover">
 			<thead>
 				<tr>
-					<th sw-rbkey="'define.firstName'"></th>
-					<th sw-rbkey="'define.lastName'"></th>
-					<th sw-rbkey="'define.email'"></th>
-					<th sw-rbkey="'define.giftMessage'"></th>
-					<th sw-rbkey="'define.quantity'"></th>
+					<th sw-rbkey="'admin.entity.orderitemgiftrecipient.firstName'"></th>
```
sumitverma: these rbkeys should be just entity.OrderItemGiftRecipient.firstName because it's an entity property and those rb key already exists. We use admin. for stuff like tabs and other admin related rbkey.

PR #3835, opened 2015-11-30
Feature workflow save issue

ian-a-hickey:
The lack of object check is what was causing the workflows to not save. Basically, its iterating through looking for anything that doesn't start with a $, and if it finds it, assumes that its a field. In this case, there is data inside (strings) that is not a field so it chokes on the form[key] thinking it has an object. So making sure we have an object before accessing its data with '.' fixes the issue.


rmarchandTen24:
Ian what is a specific example of data that exists on the form that is causing the choke?

PR #3875, opened 2015-12-14
fixing reset password process when old expired authentications exist

jasperboyd:


rmarchandTen24:
can we add a maxresults=1 to this as well

PR #3900, opened 2015-12-18
hiding load more when there are less than ten records

jasperboyd:


rmarchandTen24:
can you make this based on a variable in the scope so it scales if we change the size later

PR #3928, opened 2016-01-05
changing formatting for order not placed renewal message

jasperboyd:


rmarchandTen24:
why are we removing the else statement? if an order is not placed it won't have an order number

PR #3930, opened 2016-01-05
adding count to promotion period and promotion codes

jasperboyd:


rmarchandTen24:
don't use arraylen becaue it will only turn to sql if the property is lazy="extra"
use ​rc.promotion.getPromotionPeriodsCount()​ instead

PR #3932, opened 2016-01-05
option group count

jasperboyd:


rmarchandTen24:
array len bad. propertycount good

PR #3936, opened 2016-01-06
when adding a sku set the skucode to the product Code + skucount if a

rmarchandTen24:
sku code isn't provided

### DIFF
--- a/model/entity/Product.cfc
+++ b/model/entity/Product.cfc
@@ -1018,6 +1018,13 @@ component displayname="Product" entityname="SlatwallProduct" table="SwProduct" p
 
 	// Skus (one-to-many)
 	public void function addSku(required any sku) {
+		//if sku code is null then create one automatically
+		if(isNull(arguments.sku.getSkuCode())){
+			var skusCount = this.getSkusCount();
+			var skuCode = this.getProductCode() & "-#skusCount + 1#";
+			arguments.sku.setSkuCode(skuCode);
+		}
+		
 		arguments.sku.setProduct( this );
 	}
 	public void function removeSku(required any sku) {


model/entity/Product.cfc
```diff
@@ -1018,6 +1018,13 @@ component displayname="Product" entityname="SlatwallProduct" table="SwProduct" p
 
 	// Skus (one-to-many)
 	public void function addSku(required any sku) {
+		//if sku code is null then create one automatically
+		if(isNull(arguments.sku.getSkuCode())){
+			var skusCount = this.getSkusCount();
+			var skuCode = this.getProductCode() & "-#skusCount + 1#";
+			arguments.sku.setSkuCode(skuCode);
+		}
+
```
sumitverma: I think we should do this when creating the sku and not in the bidirectional method. Here we are just assigning the sku to product.

rmarchandTen24: I disagree because when creating the sku you will not have index info on
where the item exists in the getSkus array

gregmoser: Also this doesn't work when you delete a sku.  Consider this:

ASDF-1
ASDF-2
ASDF-3
ASDF-4
ASDF-5

Then delete ASDF-3

ASDF-1
ASDF-2
ASDF-4
ASDF-5

Now add a new one...

Your logic would make it ASDF-5... but that already exists.

rmarchandTen24: https://github.com/ten24/slatwall/pull/3938

rmarchandTen24: I just thought of that so I think the logic only should run on new instances

PR #3963, opened 2016-01-15
fixing cases where gift card number is not present calling null errors

jasperboyd:


model/service/OrderService.cfc
```diff
@@ -726,10 +726,6 @@ component extends="HibachiService" persistent="false" accessors="true" output="f
 			}
 		}
 
-		if(arguments.processObject.getSaveGiftCardToAccountFlag()){
```
jasperboyd: This logic was superfluous the gift card was already being instantiated circa lines 688 and 690

PR #3973, opened 2016-01-18
Feature team frontend directive updates

ian-a-hickey:
This brings the frontend directives branch up to date with feature. I have not until now made sure that this was all working again because I had another version to use and didn't need to upgrade anything. With this working, I should now be able to update sites that are using this and can be sure that the frontend stuff is good. 

In addition to that, there is a fix for a hibachi bug where abort was not placed correctly.


rmarchandTen24:
hey ian, just need you to resolve the conflicts

model/service/PublicService.cfc
```diff
@@ -658,6 +658,78 @@ component extends="HibachiService"  accessors="true" output="false"
...
+                shipping.setPostalCode(data.zip);
+            }
+            if (!isNull(data.phoneNumber)){
+                shipping.setPhoneNumber(data.phoneNumber);
+            }
+            if (!isNull(data.country)){
+                shipping.setCountryCode(data.country);
+            }else{
+                shipping.setCountryCode("US");
+            }
```
sumitverma: Do we need these setters? It should automatically work with populate.

ian-a-hickey: That's a good question. The answer is that I don't think that copying addresses works of of box with the populate method. Yesterday, I tried removing the setters, making sure that all the posted data was named correctly and still receive a country required error (tried with countrycode and country names) - which is odd because country is a non-persistent property of address. I think maybe we use copyAddress because setters are required for some reason with addresses - otherwise we wouldn't need it. 
https://github.com/ten24/slatwall/blob/master/model/service/AddressService.cfc#L85-L111
I had also assumed that should just work, which is why I'm passing that data in at all to the newAddress method, but maybe I shouldn't be?

ian-a-hickey: Thinking more about this - maybe we need copyAddress because we are passing in an object rather than a struct and populate only works with structs (per hibachiTransient). So that makes sense. I'm looking into why the populate doesn't work with address data now - why it is insisting on a non-persistent property being available.

sumitverma: I wasn't taking about copyAddress, I was just looking at the setters here. Instead of those you should be able to just do:
var newAddress = getService('AddressService').saveAddress(shipping,data);

sumitverma: Also, I would recommend changing the variable name to shippingAddress

ian-a-hickey: I changed the variable name to shippingAddress. Removed unneeded setters - thanks.

PR #3984, opened 2016-01-21
Feature rm custom validation

rmarchandTen24:


org/Hibachi/HibachiValidationService.cfc
```diff
@@ -31,6 +31,10 @@ component output="false" accessors="true" extends="HibachiService" {
 			}
 			// Make sure that the validation struct has contexts & properties
 			for(var customValidationKey in customValidation){
+				if(!structKeyExists(validation,'customValidationKey')){
```
sumitverma: Shouldn't customValidationKey be a variable here and not string?

PR #3989, opened 2016-01-21
adding sku codes to imagefiles method so we can have upload form for each sku image

jasperboyd:


gregmoser:
I'm confused by what this fix is.  Default Images should not be thought of as having a relationship to skuCodes.  Is there a specific task that this was done for?

jasperboyd:
The way the image form and the getDefaultImageFiles is implemented it's using skus to find its images which is where the path is stored. The bug was when you have skus with missing image files it would show multiple upload forms which if the rb keys weren't broken in them said "no images found" the fix was to display the sku code for the sku you would be uploading the image to. I was just trying to preserve what I thought was the intended functionality: https://ten24.teamworkpm.net/tasks/5026597

gregmoser:
There is a miss-understanding of how this is supposed to work.  While some products may have a unique image filename for each sku, other product configurations will have the same image filename for multiple skus.

Easiest example would be shoes that come in 10 sizes, and 3 colors.

the color option group would be set to an 'image' option group

the size option group would not.

30 total skus, with only 3 image filenames.

jasperboyd:
That's good to know, and should still work with this pull request, but saying theres no image found and not telling the user where that image is going to be attached is misleading. The case for the task is when the Sku image files are not found.

gregmoser:
I don't believe it will still work.  Line 81 on the view doesn't make any sense.  In my example 1 array would return 30 skus, the other would only return 3.  Using the same index doesn't make sense, and putting a skuCode next to a default image also doesn't make sense because there are more than 1 sku codes potentially per image.

PR #3990, opened 2016-01-21
Updated the linebreak used in custom properties to work in linux

christopher-kundrat:
enviornments


gregmoser:
Why are there 2 equals signs on 1 line, I've never known that to be valid syntax... what does

```
linebreak = linebreak = ... 
```

What does that actually mean?

PR #3993, opened 2016-01-21
Updates filter display if filter is changed

migueltarga:


migueltarga:
Pull request ins wrong branch.

PR #3994, opened 2016-01-21
fixing image count for zero case

jasperboyd:


gregmoser:
how does

```
getProductImagesCount()
```

actually work?  I don't see a property called productImages, and I don't see that function in the code.  Am I missing something?

gregmoser:
NVMD, now is see the productImages property.

-Greg

PR #4003, opened 2016-01-22
fix for frontend templates that don't have savegiftcardtoaccountflag

jasperboyd:


gregmoser:
Isn't this the exact same fix that Ryan just did merged in?  My question about just setting default=true on the property instead would apply to this as well

PR #4004, opened 2016-01-22
adding npe fix if data is missing

rmarchandTen24:


gregmoser:
Couldn't you just set the default=false on the property?

PR #4008, opened 2016-01-25
adding search to option group listing

jasperboyd:


org/Hibachi/client/src/core/components/swlistingdisplay.ts
```diff
@@ -88,8 +88,8 @@ class SWListingDisplayController{
 
     private intialSetup = () => {
         //default search is available
-        if(angular.isUndefined(this.hasSearch)){
-            this.hasSearch = true;
+        if(angular.isString(this.showSearch)){
+            this.showSearch = (this.showSearch.toLowerCase() === 'true');
```
rmarchandTen24: this should be a boolean and not a string

admin/views/entity/preprocessproduct_create.cfm
```diff
@@ -218,6 +218,7 @@ Notes:
 								data-collection="preprocessproduct_createCtrl.collection"
 								data-multiselect-field-name="options"
 								data-edit="true"
+							    data-show-search="'true'"
```
rmarchandTen24: why is this a string now?

jasperboyd:
@rmarchandTen24 changed it to a boolean

PR #4010, opened 2016-01-25
only hide pagination when expandable

jasperboyd:


gregmoser:
Why are we hiding pagination just because something is expandable?  What if the top level has more than 10 records.

-Greg

jasperboyd:
it's a limitation with the current angular listing display is what ryan just told me

PR #4034, opened 2016-01-28
Add same changes for removing whitespace from order listing as hotfix

Reyjay:
I added the same changes that I made to hotfix to features to fix the white space issue on order items view.


rmarchandTen24:
just need conflict resolution with this one

PR #4036, opened 2016-01-28
Fix header bar globally and add back to category like on hotfix

Reyjay:
This is the same change I did to hotfix.  It removes the blank ul in header and fixes categories header


gregmoser:
Making the same changes to two different branches will cause merge issues when hotfix gets merged into develop, and when feature gets merged into develop.

gregmoser:
Why wouldn't the changes just get merged from hotfix into develop or feature?

PR #4039, opened 2016-01-28
Allow overridden sku price to populate orderItem at point of adding.

chriskent:
Allow overridden sku price to populate orderItem at point of adding order item.

### DIFF
--- a/model/service/OrderService.cfc
+++ b/model/service/OrderService.cfc
@@ -380,7 +380,7 @@ component extends="HibachiService" persistent="false" accessors="true" output="f
 			newOrderItem.setSkuPrice( arguments.processObject.getSku().getPriceByCurrencyCode( arguments.order.getCurrencyCode() ) );
 
 
-			if(len(newOrderItem.getSku().getUserDefinedPriceFlag()) && newOrderItem.getSku().getUserDefinedPriceFlag() && isNumeric(arguments.processObject.getPrice()) ) {
+			if(len(newOrderItem.getSku().getUserDefinedPriceFlag()) && newOrderItem.getSku().getUserDefinedPriceFlag() && isNumeric(arguments.processObject.getPrice()) || isNumeric(arguments.processObject.getPrice()) && arguments.processObject.getPrice() neq arguments.processObject.getSku().getPriceByCurrencyCode( arguments.order.getCurrencyCode() ) ) {
 				newOrderItem.setPrice( arguments.processObject.getPrice() );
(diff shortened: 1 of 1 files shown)


model/service/OrderService.cfc
```diff
@@ -380,7 +380,7 @@ component extends="HibachiService" persistent="false" accessors="true" output="f
 			newOrderItem.setSkuPrice( arguments.processObject.getSku().getPriceByCurrencyCode( arguments.order.getCurrencyCode() ) );
 
 
-			if(len(newOrderItem.getSku().getUserDefinedPriceFlag()) && newOrderItem.getSku().getUserDefinedPriceFlag() && isNumeric(arguments.processObject.getPrice()) ) {
+			if(len(newOrderItem.getSku().getUserDefinedPriceFlag()) && newOrderItem.getSku().getUserDefinedPriceFlag() && isNumeric(arguments.processObject.getPrice()) || isNumeric(arguments.processObject.getPrice()) && arguments.processObject.getPrice() neq arguments.processObject.getSku().getPriceByCurrencyCode( arguments.order.getCurrencyCode() ) ) {
```
sumitverma: We can't allow price to be overridden by anyone, for a sku that does not allow user defined price. We will need to check  for permission of some sort. Plus, I think it will get reset when we recalculate price, if the user defined price flag is not set.

gregmoser: You should be able to override the price in the admin.  Customer Service needs to be able to do this... they were in previous version.

I don't care if we need a extra setting, or you want to make this permission based... but customer service certainly needs to be able to override the price of ANY sku (it happens all the time).

Maybe in the past there were two controllers or something, IDK.  When I originally put in the 'user defined price' it meant the user on the frontend.

sumitverma: I doubt it ever worked, but yes, I agree, it should certainly be allowed from admin. We just can't do it the way it's implemented in this pull request because it will allow price to be changed from front-end as well.

chriskent: I did a test against 3.5 and admin add item to cart price override worked

Tested the edit order item price override works.

Agree that it should be made safe to only allow as part of an admin process.

gregmoser: Yeah @chriskent you are correct, I also just tested this in 3.5, etc...  We are working on a fix and should have a hotfix in place shortly

chriskent: :)

PR #4053, opened 2016-02-02
fixing product bundle group type not saving

jasperboyd:


rmarchandTen24:
miguel's work on workflow should resolve this

PR #4059, opened 2016-02-03
render coldfusion before we read the html

rmarchandTen24:
[Review on Reviewable](https://reviewable.io/reviews/ten24/slatwall/4059)


gregmoser:
Lets talk about this, I think that this is un-needed because the rendering of CF will already be done with the include... also the way that this works potentially lends itself to race conditions.

Lastly this doesn't remove the security vulnerability, I don't think.

rmarchandTen24:
it is unnecessary and that's why it is not merged. This has nothing to do
with a security fix.

gregmoser:
Ahhh gotcha, I made an assumption.  Sorry

PR #4068, opened 2016-02-04
Updated how we are getting the values

christopher-kundrat:
for hibachiListingDisplay

### DIFF
--- a/admin/views/entity/pricegrouptabs/assignedaccounts.cfm
+++ b/admin/views/entity/pricegrouptabs/assignedaccounts.cfm
@@ -51,14 +51,8 @@ Notes:
 
 <cfparam name="rc.priceGroup" type="any" />
 
-<cfset selectedAccountDisplays = rc.priceGroup.getAccounts() />
-<cfset selectedAccountIDPaths = "" />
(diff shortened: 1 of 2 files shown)


rmarchandTen24:
@gregmoser  does the AssignedIDList method on hibachi entity in use do this in a more performant way?

gregmoser:
The previous method just caches those values within the request. see here: https://github.com/ten24/slatwall/blob/master/org/Hibachi/HibachiEntity.cfc#L381-L392

However, there is a much bigger issue with our multi-select checkbox listing displays that Adventure Cycling just recently brought up to me.  If there are let's say 10,000 accounts assigned to a price group then both the method that Chris wrote and the old method causes the server to crash if you open this page.  This is an error Adventure Cycling has seen for a while and basically just told everyone not to open the page.

This is because both methods just instantiate 10,000 accounts, loop over the ID's and put them into a list.

Now this could be fixed by converting it to a DB query, however I don't think the underlying issue will really be resolved because we're still sticking something like 10,000 UUID's into a comma separated list in the DOM.  Not very good idea.

We need to switch these multi-select listing displays to something that is angular based.

PR #4071, opened 2016-02-04
adding check for active status code when grabbing subscription order payments

jasperboyd:


model/dao/SubscriptionDAO.cfc
```diff
@@ -2,70 +2,70 @@
...
 --->
 <cfcomponent extends="HibachiDAO">
-	
+
 	<cffunction name="getUniquePreviousSubscriptionOrderPayments">
 		<cfargument name="subscriptionUsageID" type="string" required="true" />
-		
-		<cfreturn ormExecuteQuery("SELECT DISTINCT op FROM SlatwallSubscriptionOrderItem soi INNER JOIN soi.orderItem oi INNER JOIN oi.order o INNER JOIN o.orderPayments op WHERE soi.subscriptionUsage.subscriptionUsageID = :subscriptionUsageID AND op.referencedOrderPayment IS NULL", {subscriptionUsageID=arguments.subscriptionUsageID}) />
+
+		<cfreturn ormExecuteQuery("SELECT DISTINCT op FROM SlatwallSubscriptionOrderItem soi INNER JOIN soi.orderItem oi INNER JOIN oi.order o INNER JOIN o.orderPayments op WHERE soi.subscriptionUsage.subscriptionUsageID = :subscriptionUsageID AND op.referencedOrderPayment IS NULL AND activeFlag is true", {subscriptionUsageID=arguments.subscriptionUsageID}) />
```
sumitverma: order payment doesn't have active flag, it has active status. So, this should be something like this: op.orderPaymentStatusType.systemCode = 'opstActive' 

Maybe add a quick unit test to make sure the query works.

jasperboyd: I was looking at the payment method entity *facepalm I'll make this change interestingly this didn't throw an error when i tested it

jasperboyd:
tested this query directly and it is returning the desired results

PR #4079, opened 2016-02-05
Hotfix ck sync mura categories

christopher-kundrat:


sumitverma:
Please use contentService and contentDAO for all category related methods.

PR #4085, opened 2016-02-08
Allow tax exempt accounts

chriskent:


sumitverma:
Any reason why this shouldn't be just a property on account vs setting?

chriskent:
Looking at all account settings these are all done as a setting and not a property. 

Using a setting also allows setting a site default and override at account level.

sumitverma:
Big different is settings can't be used in collections. So, unless there is good usecase it might make sense to make this account property as it will make searching / reporting easier.

@gregmoser thoughts?

gregmoser:
I can see the argument on both sides.  I think we need to move away from our standard setting format in general for this issue with trying to build a query that uses it.  Maybe we only need to do this with some settings and not all.

I think a setting makes sense if we ever add an "account type", but I'm not sure that we would.

To me I also think that it makes sense as an account property so that you could set it ad-hoc on guest accounts programatically.  In fact you should probably be able to set it programatically on the order as well.

At the end of the day I think I agree with sumit.

chriskent:
OK, I will rework this and create a new account property for "tax exempt"

sumitverma:
please name the property taxExemptFlag. thanks.

Sumit Verma

gregmoser:
Lets call it taxExemptFlag

PR #4106, opened 2016-02-11
Update AttributeValue.cfc

pauaware:
We've been working with a webservice that returns a binary file, that we pre-upload from the response. The file is needed to be viewable as a custom Attribute. setAttributeValue()'s condition to check for a 'form' key has recently broken this function. We propose a mod to check for a pre-existing file path so that a File Attribute can be used in both modes.

### DIFF
--- a/model/entity/AttributeValue.cfc
+++ b/model/entity/AttributeValue.cfc
@@ -666,6 +666,9 @@ component displayname="Attribute Value" entityname="SlatwallAttributeValue" tabl
 			&& len(getAttribute().getAttributeCode())
 			&& !isNull(getAttribute().getAttributeInputType())
 			&& getAttribute().getAttributeInputType() == 'file') {
+			
+			var preExistingFilePath = getAttribute().getAttributeValueUploadDirectory() & arguments.attributeValue;
+			writeLog(file="Slatwall", text="Detail Log - FilePath > #preExistingFilePath#");
 
 			// Make sure that a new value was passed in to be set
 			if(structKeyExists(form, getAttribute().getAttributeCode()) && len(form[getAttribute().getAttributeCode()])) {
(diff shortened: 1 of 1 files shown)


gregmoser:
@rmarchandTen24 and @sumitverma what are you guys thinking about getting this PR merged in?

sumitverma:
I would prefer to change it to not use form scope at all. Entity should not
interact directly with form or url data.

Sumit Verma

gregmoser:
@sumitverma I think that is fine, but this PR isn't touching that logic.  Do you think that the rest of the changes are OK?  Do you want to open a separate issue for removing the form scope from within the entity?

As an aside, I think that we HAVE to use form scope because it's how "fileUpload()" which is the baked in CF command gets the value.  Otherwise we would have to do this in a controller WAY up the chain somewhere I guess.

sumitverma:
It's upto @rmarchandTen24 whether he wants to merge it or not, but the pull request needs to be made to hotfix, and writelog needs to get removed.

pauaware:
Hi Sumit/Greg, should I have posted this as an Issue vs PR? Cheers, Matt

gregmoser:
No, doing this as a PR is fine... we just need to get the two things that @sumitverma mentioned in place.  I'm closing this PR so that you can open a new one into 'hotfix' instead of 'master', and also remove that writeLog().

@sumitverma is it that you don't want any logging for security purposes, or would you prefer that he just change this to logSlatwal().  @pauaware we have a function called logSlatwall() that you can call from anywhere which logs to a Slatwall specific log file and allows turning on and off in the admin.

sumitverma:
Yes, using logslatwall with log level of detail is fine for these.

Sumit Verma

PR #4108, opened 2016-02-12
fix to max and min order quantity issue

rmarchandTen24:
[Review on Reviewable](https://reviewable.io/reviews/ten24/slatwall/4108)

### DIFF
--- a/model/dao/OrderDAO.cfc
+++ b/model/dao/OrderDAO.cfc
@@ -227,5 +227,76 @@ Notes:
 
 		<cfreturn 0 />
 	</cffunction>
+	<cfscript>
+		public numeric function getOrderItemCountOnOrder(required any orderItem){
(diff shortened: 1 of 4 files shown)


gregmoser:
This change is now doing double work with the orderItem lookup when the item has trackInventory turned on and the order has been placed.

Also, ormExecuteQuery() won't work if multiple of the same sku is being added to the order because the items won't be in the database yet.  You need to loop over them by calling order.getOrderItems() for the ones that have yet to persist.

I really believe that it makes more sense for these checks to exist in that master function.

rmarchandTen24:
I agree that the call needs to make a smart checks, based on if the orderitem we are operating on is persisted or not. However it is unnecessary to loop over order.getOrderITems() for every orderitem on an order. We only want to get orderitems with the same sku on the same order. This is less work overall.

PR #4111, opened 2016-02-15
Update Public service to latest

ian-a-hickey:
Adds some methods to public service.


gregmoser:
CreateAccount shouldn't log you in.  You should just chain 2 actions together because you might want to create accounts without logging people in.

Also, you should really use the getHibachiScope().flushORMSession(); function instead of just writing ormFlush() because there is logging, and it does the hasErrors check for you.

@sumitverma thoughts?

ten24git:
Yup, already mentioned these exact 2 points to Ryan. :)

PR #4113, opened 2016-02-15
Hotfix ck custom property line break

christopher-kundrat:


gregmoser:
This should be set at the Hibachi level if Hibachi is dependent on that framework setting.  Also, why aren't we just defaulting to the server scope value of the OS it's running on instead of just placing "Windows" in their by default.  This update would require all MAC / Linux installs to break until someone explicitly changes it.

gregmoser:
@rmarchandTen24 and @sumitverma maybe I'm wrong with my comment above... what are your thoughts?

christopher-kundrat:
@gregmoser we might want to change the variables.framework.environment name to something more specific but what essentially it's specify is not what your current environment is but your production environment. Basically with how it was before I was pushing up the mac LF to a windows environment which would then convert those into the windows CRLF. This was just causing problems in git. Also with how custom properties currently work, it wasn't able to recognize the difference between CRLF and LF characters so whenever you tried to run an update it wouldn't be able to recognize the current custom properties and would just rewrite them with the new linebreak causing a duplicate properties error. 

Basically this fix was to just choose one set of linebreaks to prevent any confusion between the development environment and the production environment and then allow the user to override what their production environment will be in custom.

sumitverma:
Yes, lets change the variable name to lineBreakStyle.

Sumit Verma

sumitverma:
Also, just to clarify, defaulting to windows won't break anything. That's
how code always was, there was a change made last month to use javaline
break which is now causing issues.

Sumit Verma

rmarchandTen24:
the line break should be platform independent by default. So just set the variable to "any" and the case should use resolve to the default. If you want to specify a target you can do it in custom config. @gregmoser the issue is when compiling custom properties using the java line break. It adds a character not recognized by other operating systems. So build on mac and you can't deploy on windows. We just want the option to be able to pick an OS carriage return if there is a difference

gregmoser:
I think that you guys are missing a key point of what I was saying.  Right now we are EXPLICITLY looking for variables.framework.hibachi.lineBreakStyle on line 431 in org/Hibachi/Hibach.cfc

However, we are setting that default value in the SLATWALL config of /config/configFramework.cfm which is not the right place.  It should be getting set on line 94 of /org/Hibachi/Hibachi.cfc just like all of the other defaults in Hibachi.

ALSO

You can use <cfdump var="#server.os.name#" /> to set this value so that nobody ever needs to set it manually as windows or linux, etc.

In fact the underlying function could just use this so that you NEVER have to declare a config value.

gregmoser:
Sorry, my first comment didn't output my point.  I'm just saying you can use
```
#server.os.name#
```

rmarchandTen24:
yeah that does make sense. Configuration is time consuming

PR #4115, opened 2016-02-15
filtering add sku bundle list by base product type

jasperboyd:


gregmoser:
You should be able to bundle skus from different productTypes together into 1.

jasperboyd:
https://ten24.teamworkpm.net/tasks/5284249

sumitverma:
Just closing loop on this. You can not add different product type to sku bundle. This is by design. Product bundle can be used to do that.

PR #4118, opened 2016-02-15
Removed the chained login from public service as well as the ormFlush.

ian-a-hickey:
Create account once again just creates an account. I removed the chained call to login and the ormFlush that was commented on.

### DIFF
--- a/model/service/PublicService.cfc
+++ b/model/service/PublicService.cfc
@@ -221,11 +221,10 @@ component extends="HibachiService"  accessors="true" output="false"
         param name="arguments.data.createAuthenticationFlag" default="1";
         
         var account = getAccountService().processAccount( data.$.slatwall.getAccount(), arguments.data, 'create');
-        
         getHibachiScope().addActionResult( "public:account.create", account.hasErrors() );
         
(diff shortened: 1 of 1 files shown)


model/service/PublicService.cfc
```diff
@@ -219,10 +217,13 @@ component extends="HibachiService"  accessors="true" output="false"
      *  @param password {string}
      *  @param passwordConfirm {string}
      */
-    public void function createAccount( required struct data ) {
+    public any function createAccount( required struct data ) {
         param name="arguments.data.createAuthenticationFlag" default="1";
+        
         var account = getAccountService().processAccount( data.$.slatwall.getAccount(), arguments.data, 'create');
+        
         arguments.data.$.slatwall.addActionResult( "public:account.create", account.hasErrors() );
```
rmarchandTen24: although this works it is not best practice. We shouldn't expect hibachi scope to be available on data. This still treats data as if it were rc. You can just call getHibachiScope() at the service level instead.

ian-a-hickey: I updated this to use getHibachiScope both placed in the method that it was being used.

PR #4119, opened 2016-02-15
fixes chained actions being called from the frontend.

ian-a-hickey:
This fixes chained actions not working on the frontend when using the directives. This uses a hibachiScope flushOrmSession so that actions that depend on the persisted data from a prior action can work.

### DIFF
--- a/api/controllers/public.cfc
+++ b/api/controllers/public.cfc
@@ -44,7 +44,25 @@ component accessors="true" extends="Slatwall.org.Hibachi.HibachiController"{
             rc.context = "getAccountData";
             this.get(rc);
         }else if ( StructKeyExists(arguments.rc, "context") && rc.context != "get"){
-            publicService.invokeMethod("#arguments.rc.context#", {data=arguments.rc});
+            var action = "";
+            var actions = [];
+            if (arguments.rc.context contains ","){
+                var actions = listToArray(arguments.rc.context);
+            }
+            if (!arrayLen(actions)){
+                publicService.invokeMethod("#arguments.rc.context#", {data=arguments.rc});
+            }else{
+                //iterate through all the actions calling the method.
+                for (var eachAction in actions){
+                    //Make sure there are no errors if we have multiple.
+                    if (!arguments.rc.$.slatwall.cart().hasErrors() && !arguments.rc.$.slatwall.account().hasErrors()){
+                          getHibachiScope().flushORMSession();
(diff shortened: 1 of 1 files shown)


rmarchandTen24:
@sumitverma @gregmoser I think the reason you need to use a flush is because in the other pr https://github.com/ten24/slatwall/pull/4118
you are using "actions" to run services in sequence when conversly "action" as in getHibachiScope.doAction() is a full request life cycle because it hits a controller.

gregmoser:
That is correct, when chaining contexts like this you need to maintain a list of something like:

getHibachiScope().getSuccessfulPublicContexts()

and 

getHibachiScope().getFailedPublicContexts()

Just like we do with the old-school 'actions'.  And as you mentioned with actions, you should be flushing after each context that is run so that the data is persisted.  If one of the actions fails, you abort the remainder of the chain.

I'm not sure if this PR does all that, but it should mimic public actions.

api/controllers/public.cfc
```diff
@@ -44,7 +44,25 @@ component accessors="true" extends="Slatwall.org.Hibachi.HibachiController"{
...
+            }
+            if (!arrayLen(actions)){
+                publicService.invokeMethod("#arguments.rc.context#", {data=arguments.rc});
+            }else{
+                //iterate through all the actions calling the method.
+                for (var eachAction in actions){
+                    //Make sure there are no errors if we have multiple.
+                    if (!arguments.rc.$.slatwall.cart().hasErrors() && !arguments.rc.$.slatwall.account().hasErrors()){
+                          getHibachiScope().flushORMSession();
+                          publicService.invokeMethod("#eachAction#", {data=arguments['rc']});
```
rmarchandTen24: Right now chained action flush based on the validity of the cart. If for some reason you acted on something other than the cart this would still flush. Can each action return an object that we check validity on before flushing? And if so I guess the question is if an action fails, should it break the chain?

gregmoser: Ryan, did you see my notes on the comments.  The answer is YES, any failure should break the chain and return.

gregmoser:
Basically it should mimic this...

Loop over actions (context's in this case):
https://github.com/ten24/slatwall/blob/master/meta/sample/Application.cfc#L73-L79
Notice the break; in the loop if there is a failure action.

Also notice that the "doAction" method calls an entire request lifecycle, which includes a flush by default as long as there were no errors.
https://github.com/ten24/slatwall/blob/master/org/Hibachi/Hibachi.cfc#L830-L834

PR #4127, opened 2016-02-15
Fix for credit in order cancellation

sumitverma:


gregmoser:
@sumitverma , I still think that the line in the first conditional also needs to be changed to a minus sign.  Right?  Isn't that what we said?

sumitverma:
ya, changed.

gregmoser:
Awesome, should be good now!

PR #4154, opened 2016-02-19
converting gift card to use only positive values 

jasperboyd:
working with cancel order, return order, and general order


model/service/OrderService.cfc
```diff
@@ -2402,7 +2388,7 @@ component extends="HibachiService" persistent="false" accessors="true" output="f
 			transactionType = arguments.orderPayment.getPaymentMethod().getSubscriptionRenewalTransactionType();
 		}
         if(!isNull(arguments.orderPayment.getPaymentMethod().getPaymentMethodType()) && arguments.orderPayment.getPaymentMethod().getPaymentMethodType() eq "giftCard"){
-            transactionType = arguments.orderPayment.getPaymentMethod().getPaymentMethodType();
+            transactionType = arguments.orderPayment.getOrderPaymentType().getTypeName();
         }
```
gregmoser: The null checks here no longer make sense if you're changing the value that is being used.

jasperboyd: yeah you're right I'll fix those, good catch

PR #4169, opened 2016-02-24
removing category code as required in tax category

jasperboyd:


gregmoser:
this won't work if you remove 'required' because two records with 'null' will violate the unique constraint.  There is a special validation for 'uniqueOrNull'.

gregmoser:
Take a look at attributeSetCode as an example.

jasperboyd:
Made the changes to this as the other pull. What I'm wondering is when we'd ever want to consider null when determining uniqueness?

gregmoser:
This is just how databases work.




On Wed, Feb 24, 2016 at 12:27 PM Jasper Boyd <notifications@github.com> wrote:
Made the changes to this as the other pull. What I'm wondering is when we'd ever
want to consider null when determining uniqueness?

—
Reply to this email directly or view it on GitHub .


-- Greg Moser Slatwall Product Manager | ten24 Digital Solutions http://www.slatwallcommerce.com http://www.ten24web.com
mobile | 760.518.5314
office | 877.886.5806 x107

PR #4177, opened 2016-02-25
Form Responses Listing With Pagination 

jasperboyd:


model/service/FormService.cfc
```diff
@@ -62,6 +64,22 @@ component  extends="HibachiService" accessors="true" {
...
+				value.setAttributeValueType("FormResponse");
+				if(!isNull(processObject.getNewFormQuestion().getDefaultValue())){
+					value.setAttributeValue(processObject.getNewFormQuestion().getDefaultValue());
+				} else {
+					value.setAttributeValue("");
+				}
+				this.saveFormResponse(response);
+			}
+		}
+
```
sumitverma: Do we need this logic here? I think we should just handle this on the display side. As the form responses grow over time, this could be issue.

PR #4183, opened 2016-02-27
Feature mt new listing display ui

migueltarga:


org/Hibachi/client/src/core/components/swselection.ts
```diff
@@ -47,9 +47,28 @@ class SWSelection{
                 if(!scope.name){
                     scope.name = 'selection';    
                 }
-                if(selectionService.hasSelection(scope.selectionid,scope.selection)){
-                    scope.toggleValue = true;    
+                scope.test = {};
```
migueltarga: I need to convert this directive to BindToController to fix this binding issue.

PR #4186, opened 2016-02-29
fixing arrow for root content items

jasperboyd:


org/Hibachi/client/src/core/components/expandablerecord.html
```diff
@@ -1,6 +1,6 @@
 <span style="margin-left:{{swExpandableRecord.recordDepth*40||0}}px;position:relative;padding-left: 12px;">
 	<i
-		ng-show="swExpandableRecord.collectionPromise.$$state.status"
+		ng-show="swExpandableRecord.collectionPromise.$$state.status !== 0"
```
jasperboyd: We might alternatively want to show something different for -1 (Collection failed) case

PR #4187, opened 2016-02-29
adding gift card to checkout template

jasperboyd:


gregmoser:
Why would a gift card be 'Save to Account'.  I don't really think that flags make sense.

PR #4188, opened 2016-02-29
fixing broken attribute file path on attribute code change

jasperboyd:


model/service/AttributeService.cfc
```diff
@@ -50,106 +50,122 @@ Notes:
...
 	// ====================== START: Save Overrides ===========================
-	
+
 	public any function saveAttribute(required any attribute, struct data={}) {
+
+		if(arguments.attribute.getAttributeInputType() == 'file'
+			&& structKeyExists(arguments.data, "attributeCode")
+			&& arguments.attribute.getAttributeCode() != arguments.data.attributeCode
+		){
+			for(var value in attribute.getAttributeValues()){
```
sumitverma: This loop has potential to cause issues. Let's say you use file attribute for account as profile image. If there are 50k account this will loop through 50K times when you change attribute code. Also, are we creating new folder for each code?

PR #4193, opened 2016-03-01
sw tooltip for field display

jasperboyd:


gregmoser:
@jasperboyd, will you update this branch with the latest hotfix so that our tests run?

PR #4194, opened 2016-03-01
Bug fixes and updates for swListingDisplay

migueltarga:
ability to have Hidden filters 
Added filtergroups directive:
```
<sw-listing-filter-group>
	<sw-listing-filter {...}></sw-listing-filter>
	<sw-listing-filter {...}></sw-listing-filter>
</sw-listing-filter-group>

<sw-listing-filter {...}></sw-listing-filter>
```

Radio button in listing display fixed


org/Hibachi/client/src/core/components/swlistingfiltergroup.ts
```diff
@@ -0,0 +1,34 @@
...
+
+class SWListingFilterGroupController{
+    public filters = [];
+
+    constructor(
+            public $scope,
+            public $transclude
+    ){
+            $transclude($scope,()=>{});
+            $scope.$parent.swListingDisplay.filterGroups.push(this.filters);
```
jasperboyd: You probably want this in the link function, we shouldn't be using $scope in the controller

gregmoser:
@migueltarga, will you update this branch with the latest feature so that our tests run?

PR #4199, opened 2016-03-02
Feature shipping rate price groups

ian-a-hickey:
This is tested working. I was able to set price groups on shipping rates and have my account only qualify for those rates when my account had price groups associated with it.


gregmoser:
Ian, will you update this with the latest feature so that our tests run?

PR #4204, opened 2016-03-02
fixing attribute value unit test

jasperboyd:


model/entity/AttributeValue.cfc
```diff
@@ -133,9 +133,15 @@ component displayname="Attribute Value" entityname="SlatwallAttributeValue" tabl
...
-	
+
 	public string function getPropertyTitle(){
-		return getAttribute().getAttributeSet().getAttributeSetName() & ': ' & getAttribute().getAttributeName();
+		if(!isNull(getAttribute()) && !isNull(getAttribute().getAttributeSet())){
+			return getAttribute().getAttributeSet().getAttributeSetName() & ': ' & getAttribute().getAttributeName();
+		} else if(!isNull(getAttribute())) {
+			return getAttribute().getAttributeName();
+		} else {
+			return "attributeValue";
```
jasperboyd: Not sure what the correct thing to do here is, there really shouldn't ever be an attribute value without an attribute

rmarchandTen24: @gregmoser hey greg, isn't there an rbkey you mentioned for this?

gregmoser: I would just use:

rbKey('entity.attributeValue.attributeValue')  and then make sure that rbKey actually exists.

PR #4205, opened 2016-03-02
fixing promotion service unit test

jasperboyd:


jasperboyd:
For the reason why this assertion should be one look at line 721 of the promotionService

PR #4251, opened 2016-03-09
Feature rm luceebugandemptybaseurl

rmarchandTen24:
This change is [Review on Reviewable](https://reviewable.io/reviews/ten24/slatwall/4251)


rmarchandTen24:
this is the weird part of it. the method getStatus on coldfusion exists on getPageContext().getResponse().getResponse().getStatus(). On lucee, it exists on getPageContext().getResponse().getStatus()

gregmoser:
I think in CF there are 2 places that getStatus() exists... but I have to actually test it to be sure.  It just seems really weird that there is a getResponse() inside of getResponse().  I've never really seen that before (not saying you aren't right).

PR #4260, opened 2016-03-14
adding rate name to simple representation property to display in place order

jasperboyd:


model/entity/ShippingMethodOption.cfc
```diff
@@ -117,24 +117,24 @@ component entityname="SlatwallShippingMethodOption" table="SwShippingMethodOptio
...
-	
+
 	// ===============  END: Custom Formatting Methods =====================
 
 	// ================== START: Overridden Methods ========================
-	
+
 	public any function getSimpleRepresentation() {
-		return '#getShippingMethodRate().getShippingMethod().getShippingMethodName()# - #getFormattedValue("totalChargeAfterDiscount")#';	
+		return '#getShippingMethodRate().getShippingMethod().getShippingMethodName()# - #getShippingMethodRate().getShippingMethodRateName()# - #getFormattedValue("totalChargeAfterDiscount")#';
```
rmarchandTen24: lets do
var shippingMethodRateName = "";
if(!isNull(getShippingMethodRate().getShippingMethodRateName())){
  shippingMethodRateName = getShippingMethodRate().getShippingMethodRateName();
}

and then use that in the return string to minimize null errors

gregmoser: There is also a helper function in the base hibachi object to do:

```
var shippingMethodRateName = nullReplace(getShippingMethodRate().getShippingMethodRateName(), "");
```
or just

```
return '#getShippingMethodRate().getShippingMethod().getShippingMethodName()# - #nullReplace(getShippingMethodRate().getShippingMethodRateName(), "")# - #getFormattedValue("totalChargeAfterDiscount")#'
```

so that this can all be done inline

jasperboyd: the reason I didn't do a null check here is shippingMethodRateName is a non persistent property and I'm pretty sure it will never be null, look at ShippingMethodRate.cfc line 162

PR #4263, opened 2016-03-15
fixing product bundle functionality problems

jasperboyd:


admin/client/src/productbundle/components/swproductbundlegroup.ts
```diff
@@ -325,7 +325,8 @@ class SWProductBundleGroupController {
         }
 
         //Adds filter item to designated filtergroup
-        this.productBundleGroup.data.skuCollectionConfig.filterGroups[this.index].filterGroup.push(collectionFilterItem);
+        this.productBundleGroup.data.skuCollectionConfig.filterGroups[0].filterGroup.push(collectionFilterItem);
+        this.productBundleGroup.forms[this.formName].skuCollectionConfig.$setDirty();
```
jasperboyd: @migueltarga I thought we didn't need this sort of manually dirty setting anymore

migueltarga: Yes! You should use addFilter function, never work directly with the collectionconfig filters/columns/joins...

migueltarga: And you don't need to set it Dirty anymore, (if this is the question) I think I misunderstood.

PR #4270, opened 2016-03-16
Update Content.cfc

nsollecito:
Suggestion to use CLOB here for ormtype value on urlTitlePath and contentBody. An ormtype of "string" with a length of 8000 causes issues on Oracle databases. 

Since the max length of a VARCHAR2 field on Oracle is 4000 characters, Hibernate converts this to a LONG data type which are somewhat deprecated. A table cannot contain more than one LONG column, causing the Slatwall reload to fail (silently on Railo/Lucee). 

This took a VERY long time to track down but changing the types here seems to address it. Just wanted to open this up for now as it might need further exploration / discussion.

### DIFF
--- a/model/entity/Content.cfc
+++ b/model/entity/Content.cfc
@@ -57,8 +57,8 @@ component displayname="Content" entityname="SlatwallContent" table="SwContent" p
 	property name="allowPurchaseFlag" ormtype="boolean";
 	property name="productListingPageFlag" ormtype="boolean";
 	property name="urlTitle" ormtype="string" length="4000";
-	property name="urlTitlePath" ormtype="string" length="8000";
-	property name="contentBody" ormtype="string" length="8000" ;
+	property name="urlTitlePath" ormtype="clob";
+	property name="contentBody" ormtype="clob";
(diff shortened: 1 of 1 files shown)


sumitverma:
Thanks Nick. We will check the conversion on clob in mssql and mysql. If it create varchar(max) and longtext then it's perfect solution.

nsollecito:
Thanks Sumit. Went over this with Ian this morning and I'm not sure this solution is 100% complete yet. 

Basically, I was able to get the update to work locally using ormtype="string" and length="4000". Then, after researching a bit more on ormtypes, it seems clob was the most appropriate datatype to use for an 8000 character length. However, Oracle was still using LONG in the case of updating Ian's instance. I don't know if this is a dialect-specific issue with Hibernate (wrong driver version, perhaps?)

I suggested adding a sqltype attribute to the property as well so might need to follow up here and determine the best approach that will work across MS SQL, MySQL, and Oracle.

sumitverma:
I think we will just add script to update the column based on DB. That will
work without issues across the board.

Sumit Verma

PR #4271, opened 2016-03-17
Typeahead Search Wrapper Along With Generic Directives For Collection Configs

jasperboyd:


org/Hibachi/client/src/core/components/swtypeaheadinputfield.ts
```diff
@@ -0,0 +1,102 @@
...
+	};
+	public controller=SWTypeaheadInputFieldController;
+	public controllerAs="swTypeaheadInputField";
+
+	constructor(private $hibachi, public $compile, private $timeout:ng.ITimeoutService, private utilityService, private collectionConfigService, private corePartialsPath,hibachiPathBuilder){
+		this.templateUrl = hibachiPathBuilder.buildPartialsPath(corePartialsPath) + "typeaheadinputfield.html";
+	}
+
+	public static Factory(){
+		var directive:ng.IDirectiveFactory = (
```
migueltarga: I think you only need to inject corePartialsPath and hibachiPathBuilder at this point

jasperboyd: You are right copy and pasted this from swtypeaheadsearch

rmarchandTen24:
need to decouple filter and column adding

PR #4277, opened 2016-03-18
reporting export fix

jasperboyd:
This change is [Reviewable](https://reviewable.io/reviews/ten24/slatwall/4277)


org/Hibachi/HibachiAssets/js/global.js
```diff
@@ -185,8 +185,12 @@ function initUIElements( scopeSelector ) {
 
 	jQuery.each(jQuery( scopeSelector ).find(jQuery('form')), function(index, value){
 		jQuery(value).validate({
+			submitHandler: function(form) {
+				jQuery ("button[type='submit']").prop("disabled",false);
```
rmarchandTen24: @gregmoser @jasperboyd . this will probably bring us back to where we were. I think the issue here is that submit in an export does not refresh the page as a normal submit would and therefore you cannot tap into a callback on the request to enable the button again. I think in cases such as download file or export that the button should not use submit as much as an ajax post to an iframe. That will allow us to access a success or failure to re-enable the button

jasperboyd: @rmarchandTen24 The problem I found was that because the value for the slatAction is stored on the submit button itself it was being left out of the post. The order of events with this code should be user clicks -> jquery validation runs with submit button disabled -> submitHandler is called reactivates the button to include the data with the post and posts it

PR #4282, opened 2016-03-19
Added ability to define session domain for sub-domain sharing

gregmoser:
This is important so that a site with multiple sub-domains can now set the following value in /custom/config/configFramework.cfm

```
variables.framework.hibachi.sessionCookieDomain = .mySite.com
```

This will allow for all sub-domains to use the same session for logins, etc.

Also cleaned up some misc formatting and commented out stuff.

### DIFF
--- a/org/Hibachi/Hibachi.cfc
+++ b/org/Hibachi/Hibachi.cfc
@@ -1,9 +1,9 @@
 component extends="FW1.framework" {
-	
+
 	// ======= START: ENVIORNMENT CONFIGURATION =======
 
 	// =============== configApplication
-	
+
 	// Defaults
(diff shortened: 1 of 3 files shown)


sumitverma:
I would recommend not adding a default config and setting domain, instead do a structKeyExists and set the domain. This will ensure that current behavior is not altered. 

Few things about cookie domain is, I think, they are supposed to always start with "." and they don't work on IP address. So, better to avoid any un-intended behavior.

Same thing on cfcookie function in tag service.

gregmoser:
That's funny I was thinking of something like that.  Maybe I'll just do a blank string and then check the len().

I actually originally did it that way, but then had to deal with bunch of conditionals in the HibachiTagService.cfc for the cfcookie tag.

PR #4287, opened 2016-03-21
fixing attribute bugs from form builder implementation

jasperboyd:


admin/views/entity/detailattribute.cfm
```diff
@@ -54,7 +54,7 @@ Notes:
 
 <cfparam name="rc.edit" type="boolean" />
 
-<cfif !isNull(rc.attribute.getAttributeSet())>
+<cfif structKeyExists(rc, "attributeSet")>
```
rmarchandTen24: i think you meant rc.attribute here

PR #4291, opened 2016-03-21
fixing order items not available even with valid QATS

jasperboyd:


model/entity/OrderItem.cfc
```diff
@@ -146,18 +146,16 @@ component entityname="SlatwallOrderItem" table="SwOrderItem" persistent="true" a
...
 			if(getSku().setting('skuTrackInventoryFlag') && !getSku().setting('skuAllowBackorderFlag')) {
 				if( !isNull(getStock()) && getStock().getQuantity('QATS') <= maxQTY ) {
 					maxQTY = getStock().getQuantity('QATS');
 					if(!isNull(getOrder()) && getOrder().getOrderStatusType().getSystemCode() neq 'ostNotPlaced') {
 						maxQTY += getService('orderService').getOrderItemDBQuantity( orderItemID=this.getOrderItemID() );
 					}
-				} else if(getSKU().getQuantity('QATS') <= maxQTY) {
-					maxQTY = getSku().getQuantity('QATS');
+				} else if(getSku().getQuantity('QATS') <= maxQTY) {
+					maxQTY = getSku().getCalculatedQATS();
```
rmarchandTen24: this won't work because the current sku does not get factored into the equation. By doing getCalculatedQATS you will get the last calculated QATS prior to this orderitem being added. We need to work on the result which is getSku().getQuantity('QATS')

jasperboyd: getSku().getQuantity('QATS') was returning zero for some reason even though calculatedQATS (which is what the add order item listing uses is 1000) I understand we need to use the up to date value but I'm not sure what to do with the result of getQuantity

rmarchandTen24:
you are still using calculatedQATS for validation and rbkey

gregmoser:
Guys, we shouldn't be using calculatedQATS for validation.  It was correct as qats... if there is a bug, then lets talk about why.  I'm happy to get on a call.

jasperboyd:
@gregmoser yep we're changing it, not sure to what yet it's going to need some investigation

PR #4298, opened 2016-03-24
Update order service to fix hard error

ian-a-hickey:
Error was being throw on trying to purchase a $0.00 subscription renewal product because some values were null. This pull allows the validation to appear.

---
This change is [Reviewable](https://reviewable.io/reviews/ten24/slatwall/4298)

### DIFF
--- a/model/service/OrderService.cfc
+++ b/model/service/OrderService.cfc
@@ -726,15 +726,14 @@ component extends="HibachiService" persistent="false" accessors="true" output="f
 
 		//check if the order payments paymentMethod is set to allow account to save. if true set the saveAccountPaymentMethodFlag to true
 		if (foundSubscriptionWithAutoPayFlagSet){
-			//if we have order payments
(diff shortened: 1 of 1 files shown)


model/service/OrderService.cfc
```diff
@@ -727,9 +727,9 @@ component extends="HibachiService" persistent="false" accessors="true" output="f
 		//check if the order payments paymentMethod is set to allow account to save. if true set the saveAccountPaymentMethodFlag to true
 		if (foundSubscriptionWithAutoPayFlagSet){
 			//if we have order payments
-			if (!isNull(arguments.processObject.getOrder().getOrderPayments())){
+			if (!isNull(arguments.processObject.getOrder().getOrderPayments()) && arrayLen(arguments.processObject.getOrder().getOrderPayments())){
```
sumitverma: arraylen is not needed here because the loop below won't iterate over empty array.

ian-a-hickey: I ran a test and figured out what the issue was here and we can discuss in code review.

PR #4301, opened 2016-03-24
Add testcase

ian-a-hickey:
This adds a testcase that makes sure inverse side to-many have cascade delete defined.


rmarchandTen24:
I don't think that this needs to be defined on all cases.

PR #4310, opened 2016-03-30
Hotfix team get content body by url title new

ian-a-hickey:
updated per sumit and ryans comments.

### DIFF
--- a/integrationServices/slatwallcms/BaseApplication.cfc
+++ b/integrationServices/slatwallcms/BaseApplication.cfc
@@ -227,15 +227,15 @@ component extends="Slatwall.org.Hibachi.Hibachi"{
 	}
 	
 	/** Returns the content given a urlTitle or the default content if no urlTitle is given. */
-    public any function getContentBodyByUrlTitle(urlTitlePath){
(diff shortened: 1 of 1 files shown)


integrationServices/slatwallcms/BaseApplication.cfc
```diff
@@ -226,6 +226,18 @@ component extends="Slatwall.org.Hibachi.Hibachi"{
...
+	/** Returns the content given a urlTitle or the default content if no urlTitle is given. */
+    public any function getContentByUrlTitlePath(urlTitlePath){
+        
+        var currentSite = getHibachiScope().getService('siteService').getCurrentRequestSite();
+        
+        if (!isNull(arguments.urlTitlePath) && !isNull(currentSite)){
+            var contentEntity = getHibachiScope().getService("ContentService").getContentBySiteIDAndUrlTitlePath(currentSite.getSiteID(),arguments.urlTitlePath);
+            return contentEntity;
+        }
+        return {};
```
rmarchandTen24: if you don't supply a return here. It will just return a null instead. That should be easier to check on the front end

PR #4357, opened 2016-04-14
making new attribute set display for form builder on the frontend 

jasperboyd:
This change is [Reviewable](https://reviewable.io/reviews/ten24/slatwall/4357)


tags/SlatwallAdminAttributeSetDisplay.cfm
```diff
@@ -50,16 +50,19 @@ Notes:
 <cfimport prefix="hb" taglib="../org/Hibachi/HibachiTags" />
 <cfif thisTag.executionMode is "start">
 
+	<cfif !structKeyExists(attributes, "hibachiScope")
+		  && structKeyExists(request.context, "fw")
+		  && structKeyExists(request.context.fw,"getHibachiScope")
+	>
+		<cfset attributes.hibachiScope = request.context.fw.getHibachiScope() />
+	</cfif>
+
```
jasperboyd: Can't use a default here throws a cast to java.lang.string error

rmarchandTen24: is it possible to pass hibachiscope to the tag in the first place?

On Thu, Apr 14, 2016 12:42 PM, Jasper Boyd notifications@github.com wrote:
In tags/SlatwallAdminAttributeSetDisplay.cfm :

> @@ -50,16 +50,19 @@ Notes:
>  <cfimport prefix="hb" taglib="../org/Hibachi/HibachiTags" />
>  <cfif thisTag.executionMode is "start">
>  
> +	<cfif !structKeyExists(attributes, "hibachiScope")
> +		  && structKeyExists(request.context, "fw")
> +		  && structKeyExists(request.context.fw,"getHibachiScope")
> +	>
> +		<cfset attributes.hibachiScope = request.context.fw.getHibachiScope() />
> +	</cfif>
> +


Can't use a default here throws a cast to java.lang.string error

jasperboyd: You raise a good point hibachiScope was an attribute on the tag before I started working on it, I haven't been able to find in the code where the attribute functionality adds this to Products, Orders, etc. I would guess that it's so the tag can be rendered early in the request, maybe @gregmoser  knows?

gregmoser: Perhaps... I just need a little more context around what exactly it is that we are solving.  @jasperboyd can you put 15min on my calendar next week whenever works for you to go over this?

PR #4358, opened 2016-04-14
fixing account payment detail and create

jasperboyd:
This change is [Reviewable](https://reviewable.io/reviews/ten24/slatwall/4358)


model/entity/AccountPaymentMethod.cfc
```diff
@@ -229,7 +229,12 @@ component displayname="Account Payment Method" entityname="SlatwallAccountPaymen
 	}
 
 	public void function setupEncryptedProperties() {
-		if(getCreditCardType() != "Invalid" && !isNull(getPaymentMethod()) && !isNull(getPaymentMethod().getSaveAccountPaymentMethodEncryptFlag()) && getPaymentMethod().getSaveAccountPaymentMethodEncryptFlag()) {
+		if( len(getCreditCardNumber()) > 0
```
jasperboyd: @sumitverma does this constitute as a change to how we handle credit cards?

model/validation/AccountPaymentMethod.json
```diff
@@ -23,6 +23,6 @@
 		"paymentTransactions":	[{"contexts":"delete","maxCollection":0}]
 	},
 	"populatedPropertyValidation":{
-	    "billingAddress":[{"validate":"full"}]
+	    "billingAddress":[{"conditions":"paymentTypeCreditCard","validate":"full"}]
```
gregmoser: This additional validation is no good.  If a billing address is submitted with any type of payment, then it needs to be validated in full.

migueltarga:
Reviewed 3 of 3 files at r1.
Review status: all files reviewed at latest revision, 2 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/ten24/slatwall/4358)*

model/validation/AccountPaymentMethod.json
jasperboyd: @gregmoser I'll flip the logic so it will only not check billing address if it's a gift card

gregmoser: I think that the point is... it shouldn't be checking billing address as it stands right now already... unless you pass a billingAddress form value.

PR #4362, opened 2016-04-15
Guest account fulfillment fix

Squeebly-Joe:
The getOriginalProviderTransactionID function would retrieve the a different transaction ID than the one generated by the authorize transaction for the chargePreAuthorization transaction which stopped the order from being fulfilled.

---
This change is [Reviewable](https://reviewable.io/reviews/ten24/slatwall/4362)


gregmoser:
I don't see any issues with this change.

PR #4366, opened 2016-04-18
"Select" removed from first item in typeahead

migueltarga:
This change is [Reviewable](https://reviewable.io/reviews/ten24/slatwall/4366)


gregmoser:
This change will impact existing functionality.  Be sure to look at Tools > Event Triggers in the admin... because it uses this option list as well.

PR #4376, opened 2016-04-19
Feature ih missing product image by site

ian-a-hickey:
the product image by site.

---
This change is [Reviewable](https://reviewable.io/reviews/ten24/slatwall/4376)


model/service/ImageService.cfc
```diff
@@ -118,16 +118,36 @@ component persistent="false" extends="HibachiService" output="false" accessors="
...
+			
+			//look if the path was supplied
 			if(structKeyExists(arguments, "missingImagePath") && fileExists(expandPath(arguments.missingImagePath))) {
+			
 				arguments.imagePath = arguments.missingImagePath;
+				
+		    //look if this has been supplied at the site level.
+			} else if (!isNull(getService('siteService').getCurrentRequestSite()) && !isNull(getService('siteService').getCurrentRequestSite().setting('siteMissingImagePath'))) {
+                
+                arguments.imagePath = getService('siteService').getCurrentRequestSite().setting('siteMissingImagePath');
```
gregmoser: @rmarchandTen24 I was reviewing this code by @ian-a-hickey and saw the line:

```
getService('siteService').getCurrentRequestSite()
```

Services really shouldn't know anything about requests.  getCurrentSite() should just be a property of the hibachiScope.

My architectural opinion is that hibachiScope & controllers know about request.... nothing else.  And hibachiScope is basically the request level catch-all where all information about how the request is doing, if there were errors, who the current user is, etc.... all of that request level stuff lives in the hibachiScope and it works really well because the hibachiScope is everywhere.

rmarchandTen24: I can agree with that. I'll do the spring cleaning after i merge this

PR #4381, opened 2016-04-19
[IN_PROGRESS]easy update for ps for buddies.

ian-a-hickey:
I had to fix this in Buddies so that it would stop throwing an error when trying to throw an error.

---
This change is [Reviewable](https://reviewable.io/reviews/ten24/slatwall/4381)


model/service/PublicService.cfc
```diff
@@ -995,11 +995,14 @@ component extends="HibachiService"  accessors="true" output="false"
     }
     
     public any function addErrors( required struct data , errors){
-        if (!structKeyExists(arguments.data.ajaxResponse, "errors")){
-            arguments.data.ajaxResponse["errors"] = {};
-        }
-        arguments.data.ajaxResponse["errors"] = errors;
-    } 
+    if (!structKeyExists(data, "ajaxResponse")){
```
gregmoser: This should be arguments.data

Otherwise it makes sense.

gregmoser: Actually the public service shouldn't modify the ajaxResponse directly ever.  We still need to roll out the other PubliAPI changes which address this convention issue.

PR #4384, opened 2016-04-20
Hotfx auth token with all updates working

ian-a-hickey:
This change is [Reviewable](https://reviewable.io/reviews/ten24/slatwall/4384)


model/process/Account_GenerateAuthToken.cfc
```diff
@@ -52,6 +52,9 @@ component output="false" accessors="true" extends="HibachiProcess" {
 	property name="account";
 
 	// Data Properties
+	property name="authToken";
+	property name="accessKey";
+	property name="privateAccessKey";
```
sumitverma: I don't think these properties are needed anymore.

PR #4399, opened 2016-04-26
Feature ih shipping up

ian-a-hickey:
this is the updated shipping cost surcharge with estimator.

---
This change is [Reviewable](https://reviewable.io/reviews/ten24/slatwall/4399)


gregmoser:
Is there a task for this PR.  The implementation here Isn't not exactly how I remember specing it out with you and @sumitverma

PR #4403, opened 2016-04-27
adding rounding for promotion processing

jasperboyd:
This change is [Reviewable](https://reviewable.io/reviews/ten24/slatwall/4403)

### DIFF
--- a/model/dao/PromotionDAO.cfc
+++ b/model/dao/PromotionDAO.cfc
@@ -2,58 +2,58 @@
 
     Slatwall - An Open Source eCommerce Platform
     Copyright (C) ten24, LLC
-	
+
     This program is free software: you can redistribute it and/or modify
     it under the terms of the GNU General Public License as published by
     the Free Software Foundation, either version 3 of the License, or
     (at your option) any later version.
-	
+
     This program is distributed in the hope that it will be useful,
     but WITHOUT ANY WARRANTY; without even the implied warranty of
     MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
     GNU General Public License for more details.
-	
+
     You should have received a copy of the GNU General Public License
     along with this program.  If not, see <http://www.gnu.org/licenses/>.
-    
+
     Linking this program statically or dynamically with other modules is
     making a combined work based on this program.  Thus, the terms and
     conditions of the GNU General Public License cover the whole
     combination.
-	
-    As a special exception, the copyright holders of this program give you
-    permission to combine this program with independent modules and your 
-    custom code, regardless of the license terms of these independent
-    modules, and to copy and distribute the resulting program under terms 
-    of your choice, provided that you follow these specific guidelines: 
 
-	- You also meet the terms and conditions of the license of each 
-	  independent module 
-	- You must not alter the default display of the Slatwall name or logo from  
-	  any part of the application 
-	- Your custom code must not alter or create any files inside Slatwall, 
+    As a special exception, the copyright holders of this program give you
+    permission to combine this program with independent modules and your
+    custom code, regardless of the license terms of these independent
+    modules, and to copy and distribute the resulting program under terms
+    of your choice, provided that you follow these specific guidelines:
+
+	- You also meet the terms and conditions of the license of each
+	  independent module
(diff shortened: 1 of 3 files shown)


gregmoser:
CAST() doesn't work on all DB's.  Need to verify on MS SQL, MySQL & Oracle

brucebailey:
Due to an overwhelming number of junk messages, this email address is walking among the dead.

If you are a real person with a real message for me, please resend to bailey at movementdh dot com.  

Thank you!

jasperboyd:
@gregmoser there are other cast statements in code outside of this diff, we should probably look if those are currently causing problems as well

gregmoser:
Ok, it may be fine... but you certainly need to check before merging it in.  You should be able to test MySQL and MS SQL... @ian-a-hickey can help you run it against Oracle.

jasperboyd:
@gregmoser Ian's confirmed this is okay on oracle, I've already tested on mysql and mssql and found it to fix the bug

gregmoser:
Ok, great!  Then merge away.  I must have been mistaken with CAST() being a compatibility issue.

Thanks for testing.

jasperboyd:
In some cases it would matter "CAST does not support LONG, LONG RAW, or the Oracle-supplied types." but Decimal is an Ansi SQL Type

gregmoser:
Oh ok... so I was only "mostly" wrong :)

migueltarga:
Reviewed 2 of 3 files at r1, 1 of 1 files at r2.
Review status: all files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/ten24/slatwall/4403)*

gregmoser:
Still an issue here.  CAST & ROUND don't make sense together.  Round() is a ColdFusion function, but it is being used in SQL.

Also, I don't think that you can use CAST as a replacement for ROUND, because it has different behavior... it truncates.

We should probably discuss solution as a group at this point

jasperboyd:
@gregmoser round is a sql function too

gregmoser:
Then we shouldn't need the cast right?

jasperboyd:
I was replicating the implementation of L827-828 it might be a redundancy, but it seems like a good safeguard against none currency decimal values, worst case scenario it returns what it was going to return anyway, best case it strips a .001 that could hose the whole order process

gregmoser:
Ok, well the round() right now is in the wrong spot.  the way it is right now, it's going to round() then * 100 then / 100

right now it's

```
CAST( ((round(originalPrice - (originalPrice * (discountAmount / 100)))*100)/100) AS DECIMAL(19,2) )
```

Shouldn't it be

```
CAST( ( round( ( originalPrice - (originalPrice * (discountAmount / 100) ) ) * 100 ) / 100 ) AS DECIMAL(19,2) )
```

And I still get nervous about truncation so I'd rather just see an error.  With that being said I would make it

```
round( (originalPrice - (originalPrice * (discountAmount / 100) ) ) *100 ) / 100
```

jasperboyd:
That was my mistake, I thought the first example was our convention for rounding operations.

I'm not sure what makes you nervous, if a hard error was thrown it would be one thing, but all the user sees is 'invalid payment amount', with a $0 payment amount due, It takes some writedumping to discover the rounding error.

gregmoser:
The conversation has turned purely academic at this point, but I will continue just for arguments sake.  I don't really care at this point, but I will make my points anyway.

1) If you for example do CAST(19.99999, DECIMAL(19,2)) the net result will be 19.99.  However when doing a round (as in all other promotion math) the result is 20.00

2) If you state that it will never be 19.99999 BECAUSE of the previous round, then what is the point of the CAST()?  It's just adding additional processing overhead.

jasperboyd:
I think the academia is worthwhile I wanted to understand this one, my original thought was the cast was a safeguard against the * 100 / 100 operation only because of the devision, but if that belongs in the round then I see what you're saying and the cast is superfluous. 

I was basing this implementation and understanding from L548, L584, L696, L742 which are implemented a little differently * 100 being the last operation and outside of the round. I'll make the revision you describe, do you think anything should be changed with those?

Thanks for baring with me on this one!

PR #4410, opened 2016-04-29
Final Cleanup of Access-Key, Access-Key-Secret and removal of AuthToken

gregmoser:
This change is [Reviewable](https://reviewable.io/reviews/ten24/slatwall/4410)


admin/views/entity/detailaccount.cfm
```diff
@@ -66,9 +66,9 @@ Notes:
 		<hb:HibachiEntityActionBar type="detail" object="#rc.account#" edit="#rc.edit#">
 			<hb:HibachiProcessCaller entity="#rc.account#" action="admin:entity.preprocessaccount" processContext="createPassword" type="list" modal="true" hideDisabled="false" />
 			<hb:HibachiProcessCaller entity="#rc.account#" action="admin:entity.preprocessaccount" processContext="changePassword" type="list" modal="true" />
-			<!--- If this user owns this account and this user is not a super user then allow api token generation. (if you are a super-user that wants to use the api, create a non-super-user account and set permissions.) --->
-			<cfif rc.account.getAccountID() EQ rc.$.slatwall.getAccount().getAccountID() AND rc.$.slatwall.getAccount().getSuperUserFlag() not equal "true">
-				<hb:HibachiProcessCaller entity="#rc.account#" action="admin:entity.preprocessaccount" processContext="generateAuthToken"  type="list" modal="true" />
+			<!--- If the logged in user is a super user, or they own this account then allow api token generation. --->
+			<cfif getHibachiScope().getAccount().getSuperUserFlag() || getHibachiScope().getAccount().getAccountID() eq arguments.account.getAccountID()>
```
sumitverma: this should be 

(getHibachiScope().getAccount().getSuperUserFlag() && getHibachiScope().getAccount().getAccountID() neq arguments.account.getAccountID())
||
(!getHibachiScope().getAccount().getSuperUserFlag() && getHibachiScope().getAccount().getAccountID() eq arguments.account.getAccountID())

model/entity/AccountAuthentication.cfc
```diff
@@ -2,57 +2,57 @@
...
 */
 component displayname="Account Authentication" entityname="SlatwallAccountAuthentication" table="SwAccountAuthentication" persistent="true" accessors="true" extends="HibachiEntity" cacheuse="transactional" hb_serviceName="accountService" hb_permission="account.accountAuthentications" {
-	
+
 	// Persistent Properties
 	property name="accountAuthenticationID" ormtype="string" length="32" fieldtype="id" generator="uuid" unsavedvalue="" default="";
 	property name="password" ormtype="string";
 	property name="accessKey" ormtype="string";
-	property name="authToken" ormtype="string";
+	property name="accessKeyPassword" ormtype="string";
```
sumitverma: I think we should call this accessKeyToken / accessKeySignature because it's not the password but the hashed token/signature that we are generating.

model/entity/AccountAuthentication.cfc
```diff
@@ -101,76 +101,76 @@ component displayname="Account Authentication" entityname="SlatwallAccountAuthen
...
-				if(getHibachiScope().getAccount().getAccountID() == getAccount().getAccountID() && !isNull(getAccessKey())){
-				 	var rep = "API Token - #getAuthenticationDescription()#";
-				}
-				
-		}else if(isNull(getIntegration())) {
+		if(!isNull(getAccessKey())){
+			rep &= "API Access Key - #getAuthenticationDescription()#";
+
+			// If this is a
+			if(getHibachiScope().getAccount().getSuperUserFlag() || getHibachiScope().getAccount().getAccountID() == getAccount().getAccountID()){
```
sumitverma: Same change as above. No key for super users.

sumitverma:
Same change in accountService. No authkey for super users.

admin/views/entity/detailaccount.cfm
gregmoser: So it's actually correct.  This conditional takes care of half the requirement.  The part about not being able to have an access key for a super user is taken care of in the Validation file here: model/validation/Account.json

If the account detail page you are on is a super user, that process action link never shows up in the first place because of the account validation rule (I've tested).

This logic just says... Ok, if this account IS eligible by passing validation (in other words a non-superuser)... then double check that it's either "my" account, or I'm a super user looking to create this access key for someone else's account. 

That's why this conditional matches the one in the account service.

model/entity/AccountAuthentication.cfc
gregmoser: This is just display after it's been created and it follows my same comments as above.

model/entity/AccountAuthentication.cfc
gregmoser: Couple points on this (I gave the name a LOT of thought).

1) It's not a "signature" and I don't want to confuse it with that.
2) It's not a "token" because I'm reserving the term token to mean JWTToken

I stored it in the DB as accessKeyPassword because it follows the same convention as the "password" filed.  Just like the "password" filed, we don't actually store the users password there, we store a hashed & salted value of the password.  Similarly on the accessKeyPassword we aren't actually storing the Access Key Secret (password for access key) we're storing the hashed and salted version.

If you REALLY must change it... I'm open to:

option a) AccessKeySecretHashedAndSalted (I like this more, because I'd also be willing to change the normal field to passwordHashedAndSalted)

option b) AccessKeySecret (I like this less, because it's not accurate)

PR #4411, opened 2016-04-29
typeahead multiselect peer review only

jasperboyd:
This change is [Reviewable](https://reviewable.io/reviews/ten24/slatwall/4411)


migueltarga:
Reviewed 9 of 9 files at r1.
Review status: all files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/ten24/slatwall/4411)*

org/Hibachi/client/src/core/components/swcollectionconfig.ts
```diff
@@ -65,19 +66,21 @@ class SWCollectionConfig implements ng.IDirective{
...
                 newCollectionConfig.setAllRecords(scope.swCollectionConfig.allRecords);               
                 
-                var parentScope = scope.$parent;
+                if(angular.isDefined(scope.swCollectionConfig.pageShow)){
+                    newCollectionConfig.setPageShow(scope.swCollectionConfig.pageShow); 
+                }           
                 
+                var currentScope = scope
+                 
                 for(var tries = 0; tries < 3; tries++){
```
rmarchandTen24: Hey Jasper, can you describe the problem that this loop "tries" (pun intedended) to solve :)?

org/Hibachi/client/src/core/services/typeaheadservice.ts
```diff
@@ -0,0 +1,51 @@
...
+    addRecord = (key:string, data:any) => {
+        this.typeaheadData[key] = data; 
+    } 
+    
+    getData = (key:string) => {
+        var dataToReturn = this.typeaheadData[key];
+        if(angular.isDefined(dataToReturn)){
+            return dataToReturn;
+        }
+        return false;
```
rmarchandTen24: I think that the return false isn't needed. it should return void by not having returned and that should be enough in an if statement. if(getData(key)). I know that false will evaluate the same way but now you are changing datatypes.

org/Hibachi/client/src/core/components/swcollectionconfig.ts
jasperboyd: Basically it's just inspecting for the parent directive it wants to communicate with, depending on transclusion it could be any number of levels up, 3 was the farthest I've had to go with the typeahead, but it may need to go farther the more we use this directive. I've found myself using this same loop in another directive so we may want to make this part of a service

org/Hibachi/client/src/core/services/typeaheadservice.ts
jasperboyd: Makes sense to me I'll make that change

org/Hibachi/client/src/core/components/swcollectionconfig.ts
rmarchandTen24: instead of using the try logic because you cannot predict how many parent scopes there may be, could you instead do a if(!currentScope.$parent){
var parentDirective = currentScope[scope.swCollectionConfig.parentDirectiveControllerAsName];
break;}

org/Hibachi/client/src/core/components/swclickoutside.ts
```diff
@@ -36,7 +36,7 @@ class SWClickOutside{
             if(angular.element(elem).hasClass("ng-hide")){
                 return;
             }
-            if(e.target !== elem && ! this.utilityService.isDescendantElement(elem,e.target)){
+            if(e.target !== elem && !this.utilityService.isDescendantElement(elem.context,e.target)){
```
rmarchandTen24: I see you are using elem.context. I beleive this is a deprecated method https://api.jquery.com/category/deprecated/ and if context does not exist it will default to the document dom. So first, why do we use context and how does it add value vs straight element? Pending the value it adds, I would discourage using deprecated Jquery methods

org/Hibachi/client/src/core/components/swcollectionconfig.ts
jasperboyd: sounds good!

migueltarga:
Review status: 8 of 9 files reviewed at latest revision, 4 unresolved discussions.

---

*[org/Hibachi/client/src/core/components/swtypeaheadmultiselect.ts, line 97 \[r1\]](https://reviewable.io:443/reviews/ten24/slatwall/4411#-KGwkd5G6iUEAEdIAKMA:-KGwkd5G6iUEAEdIAKMB:729157709) ([raw file](https://github.com/ten24/slatwall/blob/484f63d7a217afa72ad9906f051b90cfed74ceb9/org/Hibachi/client/src/core/components/swtypeaheadmultiselect.ts#L97)):*
white whale

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/ten24/slatwall/4411)*

org/Hibachi/client/src/core/components/swcollectionconfig.ts
jasperboyd: @rmarchandTen24 was looking at your comment again are you saying to take the for loop out? or add this conditional to the loop, I'm also already doing the equivalent of your if check at line 77.

migueltarga:
Reviewed 1 of 1 files at r2, 1 of 1 files at r3.
Review status: all files reviewed at latest revision, 4 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/ten24/slatwall/4411)*

migueltarga:
Review status: 4 of 9 files reviewed at latest revision, 5 unresolved discussions, some commit checks failed.

---

*[org/Hibachi/client/src/core/services/typeaheadservice.ts, line 51 \[r1\]](https://reviewable.io:443/reviews/ten24/slatwall/4411#-KHtiQHx2MI7rjiKZE0s:-KHtiQHx2MI7rjiKZE0t:678055703) ([raw file](https://github.com/ten24/slatwall/blob/484f63d7a217afa72ad9906f051b90cfed74ceb9/org/Hibachi/client/src/core/services/typeaheadservice.ts#L51)):*
> ```XML
> export{
>     TypeaheadService
> };
> ```

test

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/ten24/slatwall/4411)*

migueltarga:
Reviewed 5 of 5 files at r4.
Review status: all files reviewed at latest revision, 5 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/ten24/slatwall/4411)*

rmarchandTen24:
just need you to clean up the console logs.

org/Hibachi/client/src/core/components/swclickoutside.ts
```diff
@@ -36,7 +36,10 @@ class SWClickOutside{
             if(angular.element(elem).hasClass("ng-hide")){
                 return;
             }
-            if(e.target !== elem && ! this.utilityService.isDescendantElement(elem,e.target)){
+            if(e.target !== elem
```
rmarchandTen24: Since elem is sometimes and array and sometimes not. We should check if it is an array and if so then look over every element to see if a descendent is found. If it is just one element then the previouse code can be used.

PR #4425, opened 2016-05-10
fixing incorrect setting name in global advanced partial

jasperboyd:
This change is [Reviewable](https://reviewable.io/reviews/ten24/slatwall/4425)


gregmoser:
I just tested this and there is a missing RBKey

gregmoser:
@rmarchandTen24 can we have one of the interns write a unit test to make sure that all of the setting RBKeys are in place?  Maybe just loop over all the setting meta data and make sure each one resolves without a _missing.

rmarchandTen24:
@gregmoser sure will do

PR #4433, opened 2016-05-11
Hotfix team query update

ian-a-hickey:
making orderDAO Oracle safe by replacing the AS syntax with literal references.

---
This change is [Reviewable](https://reviewable.io/reviews/ten24/slatwall/4433)


rmarchandTen24:
so for oracle's sake you just need to remove the AS. I think it's still good to keep aliases mostly in the case that we have a table left joining a table with the same name like in parent child relations. This isn't one of those cases so its ok but using aliases scales better because if this query were to support that you would need to add them back anyway

<details><summary><i>Previously, ian-a-hickey (Ian Hickey) wrote…</i></summary>
> #### Hotfix team query update
> 
> making orderDAO Oracle safe by replacing the AS syntax with literal references.
> 
> 
> </details>


---

Review status: 0 of 1 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/ten24/slatwall/4433#-:-KHaZR5_7jTcDrtDjHgb:-1394935892)*

PR #4444, opened 2016-05-18
Hotfix yy query to structuretest

YuqingYoki:
The function should throw an exception when the arguments.PrimaryKey does not match any column in the query, which is shown in the third test case -- queryToStructOfStructures_notExistPrimaryKey_Test()

---
This change is [Reviewable](https://reviewable.io/reviews/ten24/slatwall/4444)


meta/tests/unit/SlatwallUnitTestBase.cfc
```diff
@@ -53,10 +53,10 @@ component extends="mxunit.framework.TestCase" output="false" {
 
 	// BEFORE ALL TESTS IN THIS SUITE
 	public void function beforeTests(){
-
+	
 		// Setup Components
 		variables.slatwallFW1Application = createObject("component", "Slatwall.Application");
-
+		variables.slatwallFW1Application.reloadApplication();
```
gregmoser: This change will slow down test running during CI because every suite of tests will do a full application reload.  Take a look at the response times on the CI server

gregmoser: I just checked the CI, and normally the tests run in ~45sec, this PR has been running for over 10 minutes (hence the timeout)

rmarchandTen24: @gregmoser, we were having issues with reloading the application via the meta/tests/RemoteFacade.cfc. @YuqingYoki  was getting an Axis java error. Have you experienced this yourself?

gregmoser: Yes, I'm pretty sure that I have.  If you guys want to setup a call and have me help debug that issue we can.

PR #4453, opened 2016-05-23
replaced getSimpleRepresentation() in orderCreateReturn

sunnysahu34:
This change is [Reviewable](https://reviewable.io/reviews/ten24/slatwall/4453)


model/process/Order_CreateReturn.cfc
```diff
@@ -83,7 +83,7 @@ component output="false" accessors="true" extends="HibachiProcess" {
 			opSmartList.addFilter('orderPaymentStatusType.systemCode', 'opstActive');
 			
 			for(var orderPayment in opSmartList.getRecords()) {
-				arrayAppend(variables.refundOrderPaymentIDOptions, {name=orderPayment.getSimpleRepresentation(), value=orderPayment.getOrderPaymentID()});
+				arrayAppend(variables.refundOrderPaymentIDOptions, {name=orderPayment.getPaymentMethod().getPaymentMethodName() & " - " & orderPayment.getCreditCardType() & " ***" & orderPayment.getCreditCardLastFour() & ' - ' & getOrder().getFormattedValue('total'), value=orderPayment.getOrderPaymentID()});
```
jasperboyd: @sunnysahu34 I think this will have problems in the case of gift card and other none credit card order payments, because getCreditCardType and getCreditCardLastFour will be null

rmarchandTen24: @gregmoser if it is a giftcard what info should we show?

gregmoser: We should just show the giftCardNumber

PR #4454, opened 2016-05-23
add default to amount type of skuPrice in orderitem.

ian-a-hickey:
This is a fix I had put into buddies. Reyjay was unable to login because he had items in his cart that were of type product bundle. Because there was no amountType set for the items, it through a hard error. This fixes it by setting a default of skuPrice to use if nothing is set.

---
This change is [Reviewable](https://reviewable.io/reviews/ten24/slatwall/4454)


model/entity/OrderItem.cfc
```diff
@@ -317,8 +317,12 @@ component entityname="SlatwallOrderItem" table="SwOrderItem" persistent="true" a
 		if(isNull(arguments.orderItem)){
 			arguments.orderItem = this;
 		}
-
-		var amountType = arguments.orderItem.getProductBundleGroup().getAmountType();
+		
+		var amountType = "skuPrice";//default
+		if (!isNull(arguments.orderItem.getProductBundleGroup())){
```
rmarchandTen24: can the default live inside of productBundleGroup.cfc property of amountType?

PR #4458, opened 2016-05-24
Workflow Fixes: Events Cache and SendFromEntityAndEmailTemplate

migueltarga:
This change is [Reviewable](https://reviewable.io/reviews/ten24/slatwall/4458)

### DIFF
--- a/model/service/WorkflowService.cfc
+++ b/model/service/WorkflowService.cfc
@@ -67,7 +67,7 @@ component extends="HibachiService" accessors="true" output="false" {
 
 				for (var workflowTrigger in workflowTriggers) {
 
-					if(workflowTrigger.getStartDateTime() > now() && (!isNull(workflowTrigger.getEndDateTime()) && workflowTrigger.getEndDateTime() < now())){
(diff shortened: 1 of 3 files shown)


model/service/WorkflowService.cfc
```diff
@@ -67,6 +67,10 @@ component extends="HibachiService" accessors="true" output="false" {
 
 				for (var workflowTrigger in workflowTriggers) {
 
+					if(workflowTrigger.getStartDateTime() > now() && (!isNull(workflowTrigger.getEndDateTime()) && workflowTrigger.getEndDateTime() < now())){
```
rmarchandTen24: if the startdate time > now and enddatetime is < now then skip. I think this may need to be an or instead of an and

gregmoser: Yeah isn't this only going to be true when the start is AFTER the end.  Shouldn't that be a validation when the trigger is setup in the first place that says the start can't be a date after the end.

migueltarga: true!! 
it should be:

`
if(workflowTrigger.getStartDateTime() > now() || (!isNull(workflowTrigger.getEndDateTime()) && workflowTrigger.getEndDateTime() < now())){
`

I will add a fix, good catch guys! Thank you

PR #4459, opened 2016-05-25
Support for  many-to-many and one-to-many filters in collections

migueltarga:
I did a typescript code format in utilityservice.ts, to fix the bad indentation everywhere

---
This change is [Reviewable](https://reviewable.io/reviews/ten24/slatwall/4459)


migueltarga:
@rmarchandTen24 since the backend gets all the default columns if we pass a blank columnsconfig, I had to create a new flag called "hasManyRelationFilter". This will trigger the loop to add groupbys in the back end instead of handling it in collectionConfigService.ts

org/Hibachi/HibachiCollectionService.cfc
```diff
@@ -451,6 +451,9 @@ component output="false" accessors="true" extends="HibachiService" {
 		if(!structKeyExists(collectionConfigStruct,'isDistinct')){
 			collectionConfigStruct.isDistinct = false;
 		}
+		if(!structKeyExists(collectionConfigStruct,'hasManyRelationFilter')){
```
rmarchandTen24: we should remove this from the collection service since the collection introspects itself

migueltarga: Done: https://github.com/ten24/slatwall/pull/4459/commits/c23c8190f8fd130947d0910481be31e0e44a049d

PR #4462, opened 2016-05-26
[IN_PROGRESS]Updates OrderDAO

ian-a-hickey:
to remove AS keyword and make oracle safe.

---
This change is [Reviewable](https://reviewable.io/reviews/ten24/slatwall/4462)

### DIFF
--- a/model/dao/OrderDAO.cfc
+++ b/model/dao/OrderDAO.cfc
@@ -136,8 +136,8 @@ Notes:
 		<cfargument name="referencedOrderID" required="true">
 
 		<cfquery name="local.giftCardOrderPayment">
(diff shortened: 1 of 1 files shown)


model/dao/OrderDAO.cfc
```diff
@@ -136,8 +136,8 @@ Notes:
 		<cfargument name="referencedOrderID" required="true">
 
 		<cfquery name="local.giftCardOrderPayment">
-			SELECT SUM(gct.debitAmount) AS amount FROM SwGiftCardTransaction AS gct
```
sumitverma: Just an FYI, it's ok to use AS in select for column alias. That should be supported on all DB types. Table alias should not use AS.

rmarchandTen24: right SUM(gct.debitAmount) AS amount should be fine.

PR #4478, opened 2016-05-28
fixing issue with coldfusion 2016

rmarchandTen24:
This change is [Reviewable](https://reviewable.io/reviews/ten24/slatwall/4478)


sumitverma:
This won't work because state code is not unique. It's supposed to be a composite key. I don't think we use postal code entity so we can just remove it.

PR #4479, opened 2016-05-31
HibachiUtilityService: add 'decimal' formatValue type and create the

YuqingYoki:
Bug happens in SKU related Price format.
* HibachiUtilityService: insert FormatValueType "decimal" and add formatValue_decimal() function;
* HibachiPropertyDisplay: Change the "currency" type to "decimal" to go into calculation without $ sign;
* Order_AddOrderItem "price" property: Add formatType as "currency" in Order_AddOrderItem "price" property.

---
This change is [Reviewable](https://reviewable.io/reviews/ten24/slatwall/4479)

### DIFF
--- a/model/process/Order_AddOrderItem.cfc
+++ b/model/process/Order_AddOrderItem.cfc
@@ -77,7 +77,7 @@ component output="false" accessors="true" extends="HibachiProcess" {
 	property name="pickupLocationID" hb_formFieldType="select" hb_rbKey="entity.orderFulfillment.pickupLocation";
 
 	// Data Properties (Inputs)
-	property name="price";
+	property name="price" hb_formatType="currency";
 	property name="currencyCode";
 	property name="quantity";
 	property name="orderItemTypeSystemCode";
--- a/org/Hibachi/HibachiTags/HibachiPropertyDisplay.cfm
+++ b/org/Hibachi/HibachiTags/HibachiPropertyDisplay.cfm
@@ -133,6 +133,10 @@
 					<cfset attributes.value = attributes.valueDefault />
 				</cfif>
 				
+				<cfif attributes.edit eq 'true' AND attributes.object.getPropertyFormatType( attributes.property ) eq 'currency'>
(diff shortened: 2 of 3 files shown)


gregmoser:
I'm not sure I fully understand this PR.  We should never be setting the 'value' of an input field to any type of 'formatted value'.  Formatted Values are just used as display and not the actual content.  Is the concern here that there is no decimal point in the UI field so it looks strange?

rmarchandTen24:
yes basically the form field renders 9.5 instead of 9.50 so formatting should be implied as it is still a ui element.

gregmoser:
I see, the other concern would be that a 'decimal' format type shouldn't always be 2 characters past the decimal point.  Do we need to add this as a format type or could we just do it in the custom tag directly?

rmarchandTen24:
this mask only back fills two decimal points of zeros. If you put in 100.555 it will still say that. However if you put 100.550 it will say 100.55. And 100.5 to 100.50. I think I would rather have it as a format type so it can be used elsewhere if needed

gregmoser:
ok

On Wed, Jun 1, 2016 at 2:13 PM Ryan Marchand <notifications@github.com>
wrote:

> this mask only back fills two decimal points of zeros. If you put in
> 100.555 it will still say that. However if you put 100.550 it will say
> 100.55. And 100.5 to 100.50. I think I would rather have it as a format
> type so it can be used elsewhere if needed
>
> —
> You are receiving this because you commented.
>
>
> Reply to this email directly, view it on GitHub
> <https://github.com/ten24/slatwall/pull/4479#issuecomment-223078867>, or mute
> the thread
> <https://github.com/notifications/unsubscribe/AAWNLM92pX0jukK4JjtjEBO7rgEqyIslks5qHcvhgaJpZM4Iq9us>
> .
>
-- 
*Greg Moser*
Slatwall Product Manager | ten24 Digital Solutions
http://www.slatwallcommerce.com
http://www.ten24web.com

mobile    |  760.518.5314
office    |  877.886.5806 x107

PR #4484, opened 2016-06-01
Prevent Auto Fulfillment went setting parameters not met

jasperboyd:
fixes problem with auto fulfillment happening on nonpayment capture

---
This change is [Reviewable](https://reviewable.io/reviews/ten24/slatwall/4484)

### DIFF
--- a/model/entity/OrderFulfillment.cfc
+++ b/model/entity/OrderFulfillment.cfc
@@ -108,7 +108,7 @@ component displayname="Order Fulfillment" entityname="SlatwallOrderFulfillment" 
 	property name="subtotalAfterDiscountsWithTax" type="array" persistent="false" hb_formatType="currency";
 	property name="taxAmount" type="numeric" persistent="false" hb_formatType="currency";
 	property name="totalShippingWeight" type="numeric" persistent="false" hb_formatType="weight";
-	property name="hasOrderWithCorrectAmountRecieved" type="boolean" persistent="false";
+	property name="hasOrderWithMinAmountRecievedRequiredForFulfillment" type="boolean" persistent="false";
 	property name="isAutoFulfillment" type="boolean" persistent="false";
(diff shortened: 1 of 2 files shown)


model/entity/OrderFulfillment.cfc
```diff
@@ -453,6 +456,30 @@ component displayname="Order Fulfillment" entityname="SlatwallOrderFulfillment"
     	return totalShippingWeight;
     }
 
+     public boolean function hasOrderWithCorrectAmountRecieved() {
```
sumitverma: I think we should change the name to hasOrderWithMinAmountRecievedRequiredForFulfillment

model/entity/OrderFulfillment.cfc
```diff
@@ -453,6 +456,30 @@ component displayname="Order Fulfillment" entityname="SlatwallOrderFulfillment"
...
+
+    public boolean function isAutoFulfillment() {
+		return (this.getFulfillmentMethodType() == "auto" || (
+		        !isNull(this.getFulfillmentMethod()) &&
+                !isNull(this.getFulfillmentMethod().getAutoFulfillFlag()) &&
+                		this.getFulfillmentMethod().getAutoFulfillFlag()));
+    }
+
+    public boolean function isAutoFulfillmentReadyToBeFulfilled(){
+		return this.isAutoFulfillment() && this.hasOrderWithCorrectAmountRecieved() && this.hasGiftCardRecipients();
```
sumitverma: This will never return true for non Gift Card orders

model/entity/OrderFulfillment.cfc
jasperboyd: sounds good

model/entity/OrderFulfillment.cfc
jasperboyd: @sumitverma It will but only because hasGiftCardRecipients returns true when it's not a gift card order, I will include that function in the refactor so it's behavior is more clear

PR #4509, opened 2016-06-09
Ability to test integrations (Only shipping integration implemeted at

sumitverma:
this time)

---
This change is [Reviewable](https://reviewable.io/reviews/ten24/slatwall/4509)


model/service/IntegrationService.cfc
```diff
@@ -299,6 +299,22 @@ component extends="HibachiService" persistent="false" accessors="true" output="f
...
 	// ===================== START: Process Methods ===========================
 	
+	public any function processIntegration_test(required any integration) {
+		var integrationTypes = getIntegrationCFC(arguments.integration).getIntegrationTypes();
+
+		for(var integrationType in integrationTypes) {
+			var integrationCFC = arguments.integration.getIntegrationCFC(integrationType);
+			if(structKeyExists(integrationCFC, "testIntegration")) {
+				var result = integrationCFC.testIntegration();
+				arguments.integration.addError(errorName="TestResult", errorMessage="#serializeJSON(result.getData())#");
```
rmarchandTen24: looks good, however, I think the only thing to be sensitive of is that the response could be either xml or json. This could be something that we declare in the variables scope of the CFC. variables.format="json".

sumitverma: Response bean is supposed to have status code (which we are not setting consistently) and data. 
https://github.com/ten24/slatwall/blob/master/model/transient/ResponseBean.cfc#L57

So, I'm just converting the struct to JSON so we can display it back in the view as message.

PR #4536, opened 2016-06-15
Add separate validation of purchaseStartDateTime and purchaseEndDateTime

YuqingYoki:
The previous code didn't consider the circumstances: If startDateTime
existed but endDateTime is null; and vice versa.

---
This change is [Reviewable](https://reviewable.io/reviews/ten24/slatwall/4536)


rmarchandTen24:
🏆

PR #4540, opened 2016-06-16
Update OrderService

ian-a-hickey:
Fixes a hard error after update when using money order. The only thing not clear to me here is this...

Does a non-persistent property, have to have an associated method? Put another way, does using 'accesors=true' work on non-persistent properties as well? I ask because it seems like we always define a method when defining a non-persistent property but I had assumed that the normal getXXX/setXXX would work for non-persistent properties as well. In which case a method will also have to be added for this and another non-persistent property in orderFulfillment.

---
This change is [Reviewable](https://reviewable.io/reviews/ten24/slatwall/4540)

### DIFF
--- a/model/service/OrderService.cfc
+++ b/model/service/OrderService.cfc
@@ -1426,7 +1426,7 @@ component extends="HibachiService" persistent="false" accessors="true" output="f
 
 		// As long as the amount received for this orderFulfillment is within the treshold of the auto fulfillment setting
 		if(
-			arguments.orderFulfillment.isAutoFulfillmentReadyToBeFulfilled()
+			arguments.orderFulfillment.getIsAutoFulfillmentReadyToBeFulfilled()
 		){
 
 			// Setup the processData


gregmoser:
@ian-a-hickey yes, getters & setters work fine.  In fact that is how the populate function works when non-persistent properties are populated during a front-end post in our process objects.  However, our convention of creating a method on an object is that the non-persistent methods on entities are mostly setup so that they can just "self-inspect" whatever they need and return the right result.  For example you shouldn't have to call

```setIsAutoFulfillmentReadyToBeFulfilled(true)```

somewhere higher up in the request just so that it can be used later.  Instead whatever validation checks you want should just existing in the getter.  It's a more Object Oriented way of programming where you don't 'tell' object's things, you 'ask' then for answers.

So in short, yes accessors work... they just get and set variables.xxx, but our convention is to create getter methods so that you aren't 'setting' a property with the exception of populate with data.  And actually in this example you wouldn't want to rely on a setter because someone on the public site could just pass "isIsAutoFulfillmentReadyToBeFulfilled" as a form value, and then take control of application behavior.

Hope this makes sense.

PR #4553, opened 2016-06-20
Hotfix ck smartlist unique flag

christopher-kundrat:
This change is [Reviewable](https://reviewable.io/reviews/ten24/slatwall/4553)


sumitverma:
What is this for? Smartlist is a list (array) and I don't think it makes sense to use it return unique object.

rmarchandTen24:
this could be useful when returning 1 record. Can we get guard rails where unique will be false if the pageShow is greater than 1 though

sumitverma:
But the purpose of smartlist is to get a list (getRecords). Also, adding unique criteria will throw error when more than one record is found. 

If you really need this, you can do something like this:

if(arraylen(variables.pageRecords) == 1 && arguments.unique == true {
return variables.pageRecords[1];
}

But even that I won't recommend, because the function can now return array OR object and you will need to check that in the code that call smartlist anyways.

rmarchandTen24:
yeah preseving the return type as array will keep the function pure and make less cases to test. I think you're right, If we want a quick function to get one record it should be a seperate function that could use the smartlist api to figure out how to render the data as unique

sumitverma:
Ya, maybe something like smartlist.getFirstRecord() because smartlist has
no way guarantee 1 unique record.

Sumit Verma

PR #4554, opened 2016-06-20
Hotfix ck add access id to create account

christopher-kundrat:
This change is [Reviewable](https://reviewable.io/reviews/ten24/slatwall/4554)


model/service/AccountService.cfc
```diff
@@ -276,6 +276,16 @@ component extends="HibachiService" accessors="true" output="false" {
 
 			arguments.account.setPrimaryEmailAddress( accountEmailAddress );
 		}
+
```
rmarchandTen24: can you just comment on your logic existing around the authentication logic. can the work be consolidated?

PR #4557, opened 2016-06-21
fixing the utility service replace string with data function 

jasperboyd:
current implementation causes bugs with typeahead, tooltip and listing

---
This change is [Reviewable](https://reviewable.io/reviews/ten24/slatwall/4557)


jasperboyd:
This change requires changes to listing display

PR #4569, opened 2016-06-23
This fixed the content listing on product pages in Sciam.

ian-a-hickey:
This change is [Reviewable](https://reviewable.io/reviews/ten24/slatwall/4569)


admin/views/entity/producttabs/listingpages.cfm
```diff
@@ -67,6 +67,8 @@ Notes:
         data-has-action-bar="false"
 		data-record-edit-action="admin:entity.editcontent"
 		data-is-angular-route="false"
+		data-get-child-count="true"
+ 		data-paginator="true"
```
rmarchandTen24: i don't think data-paginator does anything

ian-a-hickey: When you remove this.setAllRecords = true, it does not add pagination it but it only shows maybe 20 records in the list.  Adding paginator = true shows the actual paginator. Which works because it does a if(this.paginator) check. Looking through the code, I guess using showPaginator will do the same thing, but I haven't tried it. I'll check.

rmarchandTen24: so what starts my confusion is that scope on sw-listing-display doesnt' expose paginator so I assume there is no input exposed to do this. Second, I don't see any if(this.paginator) conditional in the swlistingdisplay logic so it must be an ng-if in the template. Regardless, paginator is supposed to be an object. Although setting it to true may satisfy the ng-if, it could have repercussions with the rest of the logic that expects an object.

ian-a-hickey: You are correct. Maybe it didn't build after removing the setAllRecords? Regardless, it works fine with just the pageRecords removed. Thanks for the heads-up.

rmarchandTen24: ok when you get a chance. I just need you to update the PR

PR #4586, opened 2016-06-27
Delete the function getTemplateOptions() as it never been used.

YuqingYoki:
As the title said, delete the function, and getProductTemplates() in ProductService.cfc is undefined, neither onMissing functions has related keyword

---
This change is [Reviewable](https://reviewable.io/reviews/ten24/slatwall/4586)


YuqingYoki:
Merged into wrong branch. So delete it and do a new PR.

PR #4597, opened 2016-06-30
Fixed the edit button on account page

Squeebly-Joe:
The .toggle() function has been deprecated and no worked correctly. I
replaced it with a simple function to swap element visibility. I placed
the function in the footer include. Is there maybe a better place for
it?

There is still an issue with the delete account address and payment
method functions. The page reloads, but instead of removing the address
it loads with edit form visible. Hitting cancel or save then shows the
display address.

---
This change is [Reviewable](https://reviewable.io/reviews/ten24/slatwall/4597)

### DIFF
--- a/public/views/templates/_slatwall-footer.cfm
+++ b/public/views/templates/_slatwall-footer.cfm
@@ -46,14 +46,21 @@
 Notes:
 
 --->
-	<!--- jQuery is only required if you would like to use the Slatwall client side object --->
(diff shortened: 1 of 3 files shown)


Squeebly-Joe:
Ok, never mind about the address and payment methods not deleting. The address can be deleted, but the payment method wasn't deleting because there were transactions it was being used in. None of the Slatwall error messages show by default, but that seems like one that probably should.

PR #4600, opened 2016-06-30
Develop ck avatax commit transactions

christopher-kundrat:
This change is [Reviewable](https://reviewable.io/reviews/ten24/slatwall/4600)


sumitverma:
Committing tax transaction is Avalara specific and should not be in core. All the code needs to stay in the integration.

PR #4617, opened 2016-07-07
Fix for *-to-many filters in MSSQL

migueltarga:
This change is [Reviewable](https://reviewable.io/reviews/ten24/slatwall/4617)


model/entity/Collection.cfc
```diff
@@ -851,22 +851,37 @@ component displayname="Collection" entityname="SlatwallCollection" table="SwColl
...
+					groupByList = listAppend(groupByList, collectionConfig.columns[i].propertyIdentifier);
+				}
+			}
+			if(structKeyExists(collectionConfig, 'orderBy') && arraylen(collectionConfig.orderBy) > 0){
+				for (var j = 1; j <= arraylen(collectionConfig.orderBy); j++) {
+					if (ListFindNoCase(groupByList, collectionConfig.orderBy[j].propertyIdentifier) > 0) continue;
+					groupByList = listAppend(groupByList, collectionConfig.orderBy[j].propertyIdentifier);
+				}
+			}else{
+				var baseEntityObject = getService('hibachiService').getEntityObject( getCollectionObject() );
```
rmarchandTen24: this will work. but we also have this same logic elsewhere on the getOrderByHQL function. We should make it so we don't have to maintain too different sets of if statements.

PR #4628, opened 2016-07-12
Tab Directives 

jasperboyd:
This change is [Reviewable](https://reviewable.io/reviews/ten24/slatwall/4628)


org/Hibachi/client/src/core/services/scopeservice.ts
```diff
@@ -0,0 +1,25 @@
+import {BaseService} from "./baseservice";
+
+class ScopeService { 
+
+    constructor(){
+
+    }
+
+    public locateParentScope = (scope, targetName) =>{
```
rmarchandTen24: can we change the name to getRootParentScope

jasperboyd: maybe getParentScope? it's not necessarily the root

rmarchandTen24: the root is the top most parent and it looks like the while loop keeps going until it hits the top most parent. GetParent would imply moving up one space in the chain and would not apply here.

jasperboyd: I see what you're thinking, root was just making me think of rootScope

rmarchandTen24: getTopParentScope?

PR #4629, opened 2016-07-13
Merchandising

jasperboyd:
This change is [Reviewable](https://reviewable.io/reviews/ten24/slatwall/4629)


rmarchandTen24:
Jasper, Can you change the update script to apply to 4_6 instead of 4_4? Then standard conflict resolution

rmarchandTen24:
Jasper, Can you change the update script to apply to 4_6 instead of 4_4? Then standard conflict resolution

PR #4635, opened 2016-07-19
Hotfix yy product cfc test

YuqingYoki:
Submit my final version of the ProductTest.cfc

---
This change is [Reviewable](https://reviewable.io/reviews/ten24/slatwall/4635)


meta/tests/unit/entity/ProductTest.cfc
```diff
@@ -315,4 +2240,976 @@ component extends="Slatwall.meta.tests.unit.entity.SlatwallEntityTestBase" {
...
+				}
+
+			]
+		};
+		var promotion = createPersistedTestEntity('promotion',promotionData);
+		var sku = mockProduct.getSkus()[1];
+		var promotionPeriod = promotion.getPromotionPeriods()[1];
+		var promotionReward = promotionPeriod.getPromotionRewards()[1];
+		sku.addPromotionReward(promotionReward);
+		ormFlush();
```
rmarchandTen24: although this may have gotten your test to work. It also means that the items are commited to the data base and then never cleaned up afterwards. This means that this tests data could impact the results of other tests on the dao. Ideally we shouldn't use ormflush in the tests

PR #4648, opened 2016-07-21
Hotfix yy test gettypebysystemcode function in type service and dao

YuqingYoki:
This function is called several times from Order.cfc, so I test it first.

---
This change is [Reviewable](https://reviewable.io/reviews/ten24/slatwall/4648)


model/service/TypeService.cfc
```diff
@@ -64,6 +64,7 @@ component extends="HibachiService" accessors="true" output="false" {
 		return getTypeDAO().getSystemCodeTypeCount(argumentCollection=arguments);
 	}
 	
+	//@Suppress
```
rmarchandTen24: looks good, just need to update the annotation to multiline comment

PR #4650, opened 2016-07-21
Send24 integration

migueltarga:
This change is [Reviewable](https://reviewable.io/reviews/ten24/slatwall/4650)

### DIFF
--- a/integrationServices/send24/model/service/send24Service.cfc
+++ b/integrationServices/send24/model/service/send24Service.cfc
@@ -51,10 +51,12 @@ component extends="Slatwall.model.service.HibachiService" accessors="true" outpu
 	property any integration;
 
 	public any function init(){
+
 		variables.integration = getService("integrationService").getIntegrationByIntegrationPackage("send24");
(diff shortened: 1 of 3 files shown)


integrationServices/send24/model/service/send24Service.cfc
```diff
@@ -0,0 +1,201 @@
...
+	}
+
+	private string function getSend24Auth(){
+
+		return 'Basic '&ToBase64(integration.setting('basicAuthUser') & ":" & integration.setting('basicAuthPassword'));
+	}
+
+	private string function buildSend24URL(required string endpoint){
+
+		var url ='http://' & integration.setting('companyName');
```
rmarchandTen24: this should be urlstring instead of url to avoid an error lucee throws where it can't differentiate between url the variable and url the scope

rmarchandTen24: @sumitverma should the protocol be http? https? or configurable via setting?

PR #4682, opened 2016-07-28
fixing export form responses throwing invalid token

jasperboyd:
This change is [Reviewable](https://reviewable.io/reviews/ten24/slatwall/4682)

### DIFF
--- a/admin/client/src/formbuilder/components/swformresponselisting.ts
+++ b/admin/client/src/formbuilder/components/swformresponselisting.ts
@@ -34,14 +34,21 @@ class SWFormResponseListingController {
     }
     
     export = () => {
-        $('body').append('<form action="' 
-              + this.$hibachi.getUrlWithActionPrefix() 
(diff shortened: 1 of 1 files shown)


admin/client/src/formbuilder/components/swformresponselisting.ts
```diff
@@ -34,14 +34,21 @@ class SWFormResponseListingController {
...
-        $('body').append('<form action="' 
-              + this.$hibachi.getUrlWithActionPrefix() 
-              + 'api:main.exportformresponses&formID=' + this.formId  
-              + '" method="post" id="formExport"></form>');
-              
-        $('#formExport')
-            .submit()
-            .remove();
+
+        var exportFormResponsesPromise = this.$http({
```
rmarchandTen24: can you update this to use admin request from the request service?

jasperboyd: yep yep

PR #4698, opened 2016-08-02
Update Shipping Service

ian-a-hickey:
This adds some checks to make sure that the shipping method options that are being added are allowed to be added. Previously, it was adding both manual and worldwide shipping methods to the fulfillment that it should not have been. Also, I updated the condition that decides when a new shipping method rate should be selected by default. Before, it would auto select if one and only one was available. I changed it to auto select the first one if at least one is available.

---
This change is [Reviewable](https://reviewable.io/reviews/ten24/slatwall/4698)

### DIFF
--- a/model/service/ShippingService.cfc
+++ b/model/service/ShippingService.cfc
@@ -165,6 +165,7 @@ component extends="HibachiService" persistent="false" accessors="true" output="f
 			var shippingMethodRate = shippingMethodRates[r];
 			// If this rate is a manual one, then use the default amount
 			if(isNull(shippingMethodRate.getShippingIntegration())) {
+				
 				var shipmentItemMultiplier = 0;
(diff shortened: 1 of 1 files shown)


gregmoser:
@rmarchandTen24 and @sumitverma this is going to break a lot of backward compatibility.  We can't have Slatwall just auto-selecting a shipping method for users.

sumitverma:
Where do you see that change? It only selects if there is only one method.

gregmoser:
I don't know, I just read @ian-a-hickey comments.

ten24git:
Ya, comment is incorrect.

Sumit Verma

gregmoser:
Well the the comment in the code should be switched back to what it was... because this leaves the comments in the code out of sync with it's behavior.

PR #4708, opened 2016-08-05
making comments editable for the account that commented

jasperboyd:
This change is [Reviewable](https://reviewable.io/reviews/ten24/slatwall/4708)


model/entity/Comment.cfc
```diff
@@ -70,6 +70,7 @@ component displayname="Comment" entityname="SlatwallComment" table="SwComment" p
 	// Non-Persistent Properties
 	property name="primaryRelationship" persistent="false";
 	property name="commentWithLinks" persistent="false";
+	property name="canEdit" persistent="false";
```
rmarchandTen24: unless you are populating variables.canEdit. there is no need to create a property. Because there is no need to first level cache this as the boolean should evaluate every time you can remove the property

PR #4710, opened 2016-08-05
Only consider root order items for order fulfillment subtotal

jasperboyd:
This change is [Reviewable](https://reviewable.io/reviews/ten24/slatwall/4710)


model/entity/OrderFulfillment.cfc
```diff
@@ -433,7 +433,9 @@ component displayname="Order Fulfillment" entityname="SlatwallOrderFulfillment"
   		if( !structKeyExists(variables,"subtotal") ) {
 	    	variables.subtotal = 0;
 	    	for( var i=1; i<=arrayLen(getOrderFulfillmentItems()); i++ ) {
-	    		variables.subtotal = precisionEvaluate(variables.subtotal + getOrderFulfillmentItems()[i].getExtendedPrice());
+	    	    if(isNull(getOrderFulfillmentItems()[i].isRootOrderItem())){
```
rmarchandTen24: isRootOrderItem, since it is an IS function, should always return true or false IMO. I think it is unnecessary to make it a nonpersistent property. Just make it a function, besides if the data type is boolean you would get an error anyway if it were ever null.

jasperboyd: @rmarchandTen24 yeah the isnull is a remnent from seperating out the function nice catch!

PR #4717, opened 2016-08-08
Hotfix ss schedule test

sunnysahu34:
This change is [Reviewable](https://reviewable.io/reviews/ten24/slatwall/4717)

### DIFF
--- a/meta/tests/unit/entity/ScheduleTest.cfc
+++ b/meta/tests/unit/entity/ScheduleTest.cfc
@@ -1,0 +1,107 @@
+/*
+
+    Slatwall - An Open Source eCommerce Platform
+    Copyright (C) ten24, LLC
+
+    This program is free software: you can redistribute it and/or modify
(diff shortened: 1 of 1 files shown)


meta/tests/unit/entity/ScheduleTest.cfc
```diff
@@ -0,0 +1,118 @@
...
+	public void function getRecuringTypeOptionsTest()
+	{
+		   var scheduleData={
+		   	scheduleID=''
+		                    };
+		    
+		var mockSchedule = createTestEntity('Schedule',scheduleData);
+		var result= mockSchedule.getRecuringTypeOptions();
+	
+		assertEquals("Daily", result[1].name);
```
rmarchandTen24: so we don't really want to assert that Daily is the first item in the list. because if someone adds a new item it will fail the test for an inappropriate reason. If we can get it to assert that daily just exists in the array agnostic of the position, that would be a better test

PR #4720, opened 2016-08-09
Allowing for Address to be picked when fulfillment is in order req list

jasperboyd:
In addition I refactored rc.placeOrderNeedsFulifllmentCharge to have the correct spelling for fulfillment

---
This change is [Reviewable](https://reviewable.io/reviews/ten24/slatwall/4720)


admin/views/entity/preprocessorder_placeorder.cfm
```diff
@@ -70,7 +70,7 @@ Notes:
 				<cfif listFindNoCase(rc.order.getOrderRequirementsList(), 'fulfillment')>
 					<cfset ofIndex=0 />
 
-					<cfloop array="#rc.order.getOrderFulfillments()#" index="orderFulfillment">
+					<cfloop array="#rc.order.getOrderFulfillments()#" index="local.orderFulfillment">
```
rmarchandTen24: you don't need to use local in a .cfm and i think it will actually break this unless all references go to local

PR #4727, opened 2016-08-11
if a product is marked inactive when it used to be active then set all

rmarchandTen24:
skus inactive and published false

---
This change is [Reviewable](https://reviewable.io/reviews/ten24/slatwall/4727)

### DIFF
--- a/model/dao/ProductDAO.cfc
+++ b/model/dao/ProductDAO.cfc
@@ -57,7 +57,7 @@ Notes:
 			var updateQuery = new Query();
 			var sql = '
 				UPDATE SwSku s
-				SET s.activeFlag=0,publishedFlag=0
+				SET s.activeFlag=0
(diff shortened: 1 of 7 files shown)


model/dao/ProductDAO.cfc
```diff
@@ -49,6 +49,21 @@ Notes:
...
 	<cfscript>
+		public void function setSkusAsInactiveByProduct(required any product){
+			setSkusAsInactiveByProductID(arguments.product.getProductID());
+		}
+		
+		public void function setSkusAsInactiveByProductID(required string productID){
+			var updateQuery = new Query();
+			var sql = '
+				UPDATE SwSku s
+				SET s.activeFlag=0,publishedFlag=0
```
sumitverma: sku doesn't have publishedFlag

PR #4737, opened 2016-08-15
all amazon sessions

ian-a-hickey:
This change is [Reviewable](https://reviewable.io/reviews/ten24/slatwall/4737)


model/transient/HibachiScope.cfc
```diff
@@ -107,10 +107,7 @@ component output="false" accessors="true" extends="Slatwall.org.Hibachi.HibachiS
 	}
 	
 	public boolean function getLoggedInFlag() {
-		if(!getSession().getAccount().getNewFlag() && !getSession().getAccount().getGuestAccountFlag()) {
-			return true;
-		}
-		return false;
+		return getSession().getLoggedInFlag();
```
rmarchandTen24: need to add  && !getSession().getAccount().getGuestAccountFlag()

org/Hibachi/HibachiSessionService.cfc
```diff
@@ -109,49 +123,54 @@ component output="false" accessors="true" extends="HibachiService"  {
...
 		
 		// Check to see if this session has an accountAuthentication, if it does then we need to verify that the authentication shouldn't be auto logged out
 		
 		// If there was an integration, then check the verify method for any custom auto-logout logic
 		
-		// If the sessions account is and admin and last request by the session was 15 min or longer ago. 
+		// If the sessions account is an admin and last request by the session was 15 min or longer ago. 
 		
-		if((getHibachiScope().getSessionFoundPSIDCookieFlag() && getHibachiScope().getLoggedInFlag())
+		if( (getHibachiScope().getSessionFoundExtendedPSIDCookieFlag() || getHibachiScope().getSessionFoundPSIDCookieFlag())  && !getHibachiScope().getSession().getLoggedInFlag()
```
rmarchandTen24: !getHibachiScope().getSession().getLoggedInFlag() changed to !getHibachiScope().getLoggedInFlag()

org/Hibachi/HibachiSessionService.cfc
```diff
@@ -192,20 +224,40 @@ component output="false" accessors="true" extends="HibachiService"  {
...
 			auditLogData.account = currentSession.getAccount();
 		}
-	
-		currentSession.removeAccount();
-		currentSession.removeAccountAuthentication();
-	
+		
+		//No need to remove the account or authentication. We just set the state to being logged out.
+		currentSession.setLoggedOutDateTime(DateTimeFormat(now()));
+		currentSession.setLoggedInFlag(false);
```
rmarchandTen24: remove

model/entity/Session.cfc
```diff
@@ -52,11 +52,15 @@ component displayname="Session" entityname="SlatwallSession" table="SwSession" p
 	property name="sessionID" ormtype="string" length="32" fieldtype="id" generator="uuid" unsavedvalue="" default="";
 	property name="shippingAddressPostalCode" ormtype="string";
 	property name="lastRequestDateTime" ormtype="timestamp";
+	property name="loggedInDateTime" ormtype="timestamp";
+	property name="loggedOutDateTime" ormtype="timestamp";
+	property name="loggedInFlag" ormtype="boolean";
```
rmarchandTen24: remove

admin/layouts/default.cfm
```diff
@@ -101,9 +101,10 @@ Notes:
 	</head>
 
 	<!--- Start old navbar --->
-	<body <cfif !$.slatwall.getLoggedInAsAdminFlag() && !structKeyExists(url,'ng')>class="s-login-screen"</cfif>>
+	<body <cfif (!$.slatwall.getLoggedInFlag() || !$.slatwall.getLoggedInAsAdminFlag()) && !structKeyExists(url,'ng')>class="s-login-screen"</cfif>>
```
gregmoser: I'm just confused by why this was added.  Before these conditionals were checking to make sure that an "admin" account was logged in and now you've changed it so that any account will trigger this view.

admin/views/main/default.cfm
```diff
@@ -52,6 +52,7 @@ Notes:
 <cfparam name="rc.productReviewSmartList" type="any" />
 
 <cfoutput>
+	<cfif $.slatwall.getLoggedInFlag()>
```
gregmoser: Again, not sure why this was just added to the view if it was never needed before.  The admin subsystem already has permission checking and redirects in place.

model/entity/Session.cfc
```diff
@@ -71,14 +74,71 @@ component displayname="Session" entityname="SlatwallSession" table="SwSession" p
...
+		return true;
+		
+	} 
+	
+	/** Because we are never removing the account from the session - 
+	    it becomes important that if we are returning this data, we only do that 
+	    if the user is not an admin user that is not logged in, or for a public user
+	    only if they are in the extended session period or earlier. 
+	*/
+	public any function getAccount(boolean testAccount=false) {
```
gregmoser: There shouldn't be any "testAccount" arguments on entities, the model shouldn't have any testing functionality built in.

meta/tests/unit/SlatwallUnitTestBase.cfc
```diff
@@ -63,8 +63,13 @@ component extends="mxunit.framework.TestCase" output="false" {
 	// BEFORE EACH TEST
 	public void function setUp() {
 		variables.slatwallFW1Application.bootstrap();
-
-		request.slatwallScope.getAccount().setSuperUserFlag(1);
```
gregmoser: slatwallScope.getAccount() already automatically returns a newAccount if one doesn't exists.  If you didn't do getSession().getAccount() you wouldn't need the getAccount(**TRUE**), and could remove the testAccount argument.

org/Hibachi/HibachiSessionService.cfc
```diff
@@ -162,22 +190,15 @@ component output="false" accessors="true" extends="HibachiService"  {
...
 	public string function loginAccount(required any account, required any accountAuthentication) {
 	
 		var currentSession = getHibachiScope().getSession();
 		currentSession.setAccount( arguments.account );
 		currentSession.setAccountAuthentication( arguments.accountAuthentication );
-	    
+	    currentSession.setLoggedInDateTime(DateTimeFormat(now()));
+		
 		// Make sure that we persist the session
 		persistSession();
```
gregmoser: @ian-a-hickey because persistSession() is explicitly called on every login... that entire block that you moved out of the persistSession() function and down below here now can go back into the persistSessions() function.

I also like that the function of setting the sessionID and all the new cookie values are left in that abstracted function.  I've called that function in custom code as well.

org/Hibachi/HibachiSessionService.cfc
```diff
@@ -189,23 +210,63 @@ component output="false" accessors="true" extends="HibachiService"  {
...
+	    var cookieValue = getValueForCookie();
+			getHibachiScope().getSession().setSessionCookiePSID(cookieValue);
+			getHibachiTagService().cfcookie(name="#getApplicationValue('applicationKey')#-PSID", value=getHibachiScope().getSession().getSessionCookiePSID(), expires="never");
+		
+		//only set this if the use is not an admin user and we are using extended sessions.
+		if (!getHibachiScope().getAccount().getAdminAccountFlag() && getHibachiScope().setting('globalExtendedSessionAutoLogoutInDays') && getHibachiScope().setting('globalUseExtendedSession')){
+			var cookieValue = getValueForCookie();
+			getHibachiScope().getSession().setSessionCookieExtendedPSID(cookieValue);
+			getHibachiTagService().cfcookie(name="#getApplicationValue('applicationKey')#-ExtendedPSID", value=getHibachiScope().getSession().getSessionCookieExtendedPSID(), expires="#getHibachiScope().setting('globalExtendedSessionAutoLogoutInDays')#");
+		}
```
gregmoser: These three settings of the cookies above here are what I was talking about moving back into persistSession()

org/Hibachi/HibachiSessionService.cfc
gregmoser: Actually, we can also move the getHibachiDAO().flushORMSession(); call into persistSession(), then that entire function will be very expressive... it updates the database and the cookies that relate to the database.

admin/layouts/default.cfm
ian-a-hickey: All this is doing is setting the login screen class. I believe that the login screen class should show both when you are not logged in, and when you are not logged in as admin. I could be wrong but it made sense at the time.

org/Hibachi/HibachiSessionService.cfc
ian-a-hickey: I reverted these into persistSession along with the call to persist the session. Makes sense and should still function the same way.

admin/views/main/default.cfm
ian-a-hickey: I think I added this because previously, when we were removing the account from the session, same request redirects didn't matter because there was no account on which to get information. Now, say you logout explicitly - great. That is fine. Say you timeout and it needs to redirect to the login screen...Because it redirects in place, in the same request, the account is still attached because it hasn't gone through a page refresh to delete the cookies. So even though on the next reload of the page, the cookies will be gone and you are logged out, I was finding that redirects that happened within the same request still had access. So I added an explicit check on getLoggedIn which will return false even before it refreshes, which fixed my problem.

org/Hibachi/HibachiSessionService.cfc
```diff
@@ -109,103 +123,151 @@ component output="false" accessors="true" extends="HibachiService"  {
...
 		
 		// Check to see if this session has an accountAuthentication, if it does then we need to verify that the authentication shouldn't be auto logged out
 		
 		// If there was an integration, then check the verify method for any custom auto-logout logic
 		
-		// If the sessions account is and admin and last request by the session was 15 min or longer ago. 
+		// If the sessions account is an admin and last request by the session was 15 min or longer ago. 
 		
-		if((getHibachiScope().getSessionFoundPSIDCookieFlag() && getHibachiScope().getLoggedInFlag())
+		if((getHibachiScope().getSessionFoundPSIDCookieFlag()||getHibachiScope().getSessionFoundExtendedPSIDCookieFlag()||getHibachiScope().getSessionFoundNPSIDCookieFlag() && !getHibachiScope().getLoggedInFlag())
```
gregmoser: Need parens fix

gregmoser: Also, don't need all conditionals here because the getHibachiScope().getLoggedInFlag() now does the time check

gregmoser: Move bottom two checks from hibachiScope to session service for getLoggedInFlag

org/Hibachi/HibachiSessionService.cfc
```diff
@@ -109,103 +123,151 @@ component output="false" accessors="true" extends="HibachiService"  {
...
+		
 		getHibachiAuditService().logAccountActivity( "login", auditLogData );
 		getHibachiEventService().announceEvent("onSessionAccountLogin");
-	
+		
 	}
 	
-	public void function logoutAccount() {
+	/** Logs out the user completely. */
+	public void function logoutAccount(softLogout=false) {
```
gregmoser: boolean softLogout=false

gregmoser:
@sumitverma, @rmarchandTen24, @ian-a-hickey this should not have been merged in yet.  There are still a couple of issues.  Notably placing the time check still inside of the session object, but we still need to review again @ian-a-hickey

gregmoser:
Also the !getSession().getAccount().getGuestAccountFlag() as part of getLoggedInFlag() in hibachiScope doesn't need to be there anymore.

PR #4743, opened 2016-08-17
adding redeem gift card to account process in detail account

jasperboyd:
This change is [Reviewable](https://reviewable.io/reviews/ten24/slatwall/4743)


gregmoser:
@rmarchandTen24 I thought we agreed to change this from 'redeem' to 'attach' before it got merged in.  Do you agree @sumitverma & @jasperboyd

jasperboyd:
I'm standing by my original wording for the function, redeem implies gaining of value, when a gift card is attached to account that account gains the implicit value of the gift card

gregmoser:
I appreciate that @jasperboyd, but it's not your decision.  The term "Redeem" needs to follow the typical vernacular for Gift Card processing that is an industry standard and means to "USE" the gift card for a purchase, or convert it to a cash value.  We can't just make up what the term means when the entire retail industry already uses the ther "Redeem" to mean something else.

@sumitverma I thought that the 2 of us had reached a consensus on this.

gregmoser:
Just to provide further clarification... If this process method was taking the gift card balance and converting it to an account credit balance (or term account balance in other words) then having the process be "Account - Redeem Gift Card" would be completley accurate.  That is exactly the process that Amazon outlines (https://www.amazon.com/gp/help/customer/display.html?nodeId=201138870).  Keep in mind however even if you look at the steps in that example the button they tell you to press is "Apply a Gift Card to Your Account".

Now what amazon is describing isn't even what is happening here if I understand correctly.  We aren't even applying the gift card to the account balance so that it can be used on the next order.  We are just marking this gift card as only being usable by that account.

PR #4776, opened 2016-08-24
Hotfix ss location test

sunnysahu34:
This change is [Reviewable](https://reviewable.io/reviews/ten24/slatwall/4776)

### DIFF
--- a/meta/tests/unit/entity/LocationTest.cfc
+++ b/meta/tests/unit/entity/LocationTest.cfc
@@ -76,5 +76,36 @@ component extends="Slatwall.meta.tests.unit.entity.SlatwallEntityTestBase" {
 	     
 		    assertTrue(result);
 	}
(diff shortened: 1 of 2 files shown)


meta/tests/unit/entity/LocationTest.cfc
```diff
@@ -0,0 +1,228 @@
...
+	    var mockLocation = createPersistedTestEntity('Location',locationData);
+	    
+	    var result= mockLocation.getBaseLocation();
+	    
+	    var outputExpected= "sunny";
+	    
+	    assertEquals(outputExpected, result);
+	}
+	
+	public void function isDeletableTest()
```
rmarchandTen24: can we also do a test where the location is not deletable? You should be able to do this by looking at the location.json validation rules

PR #4779, opened 2016-08-24
Adding Manual Balance Adjust for Gift Card with History Page Improvements

jasperboyd:
This change is [Reviewable](https://reviewable.io/reviews/ten24/slatwall/4779)


model/process/GiftCard_AddCredit.cfc
```diff
@@ -58,4 +58,19 @@ component output="false" accessors="true" extends="HibachiProcess"{
...
 	 }
-}
\ No newline at end of file
+
+    public any function getOrderPayments(){
+        if(structKeyExists(variables, "orderPayments")){
+            return variables.orderPayments; 
+        }
+        return []; 
+    }
```
sumitverma: Here is how we handle this in HibachiEntity, would recommend doing the same for HibachiProcess:
https://github.com/ten24/slatwall/blob/master/org/Hibachi/HibachiEntity.cfc#L18-L36

model/validation/GiftCard_OfflineTransaction.json
```diff
@@ -0,0 +1,5 @@
+{
+    "properties":{
+        "amount" : [{"dataType":"numeric", "gt":0}]
+    }
```
sumitverma: We should add the validation to make sure amount doesn't exceed the balance

PR #4789, opened 2016-08-25
ObserverService Events not case sensitive

migueltarga:
This change is [Reviewable](https://reviewable.io/reviews/ten24/slatwall/4789)

### DIFF
--- a/org/Hibachi/client/src/core/services/observerservice.ts
+++ b/org/Hibachi/client/src/core/services/observerservice.ts
@@ -37,10 +37,11 @@ class ObserverService extends BaseService{
      * @description adds events listeners
      */
     attach = (callback:any, event:string, id?:string):void => {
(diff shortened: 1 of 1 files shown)


org/Hibachi/client/src/core/services/observerservice.ts
```diff
@@ -37,18 +37,19 @@ class ObserverService extends BaseService{
      * @description adds events listeners
      */
     attach = (callback:any, event:string, id?:string):void => {
+        event = event.toLowerCase();
```
rmarchandTen24: can we apply the same lowercasing to the id as well?

migueltarga: done:
https://github.com/ten24/slatwall/pull/4789/commits/8987864dfade58aa5534c73deeb417f72ff45cee

PR #4795, opened 2016-08-25
small improvements to collection config service see notes: 

jasperboyd:
ability to define keywordFilterGroups and columns which allow developers to provide different filters and columns based on the presence of a keyword or not, also the ability to remove filters

---

This change is [Reviewable](https://reviewable.io/reviews/ten24/slatwall/4795)


org/Hibachi/client/src/collection/services/collectionconfigservice.ts
```diff
@@ -185,9 +189,19 @@ class CollectionConfig {
 
     public getOptions= (): Object =>{
         this.validateFilter(this.filterGroups);
+        if(this.keywords.length && this.keywordColumns.length > 0){
```
rmarchandTen24: i think we want to support mixing keywords with columns and filters rather than choosing either or

PR #4811, opened 2016-08-31
adding override to getskussmartlist to add an order by on sku code

jasperboyd:
This change is [Reviewable](https://reviewable.io/reviews/ten24/slatwall/4811)


model/entity/Product.cfc
```diff
@@ -217,6 +217,13 @@ component displayname="Product" entityname="SlatwallProduct" table="SwProduct" p
 		return count;
 	}
 
+	public any function getSkusSmartList(){
```
rmarchandTen24: @sumitverma @gregmoser. Would this order by apply to 90% of the situations when calling product.getSkusSmartlist()?

sumitverma: I think it should be ok, since by default there was no sortorder.

gregmoser: Why don't we override this in the service?  That way anywhere that a sku smart list comes from will automatically be sorted by sku-code.

gregmoser: Just as an FYI, when this function isn't defined the smartList still comes from the skuService where it has been overwritten with an order defined.  Not sure if that matters to you @sumitverma

jasperboyd: @gregmoser that makes more sense I didn't realize they were both getting there smart list from the service, I'll update the pull request

PR #4827, opened 2016-09-06
Hotfix ck company payment method flag

christopher-kundrat:
This change is [Reviewable](https://reviewable.io/reviews/ten24/slatwall/4827)


rmarchandTen24:
So this is pretty simple. Is there any backend logic that the flag will drive in a future PR?

PR #4844, opened 2016-09-21
Created smartlist function for getFirst record which only returns one

christopher-kundrat:
object instead of an array.

---
This change is [Reviewable](https://reviewable.io/reviews/ten24/slatwall/4844)

### DIFF
--- a/org/Hibachi/HibachiSmartList.cfc
+++ b/org/Hibachi/HibachiSmartList.cfc
@@ -810,6 +810,14 @@ component accessors="true" persistent="false" output="false" extends="HibachiObj
 		return variables.pageRecords;
 	}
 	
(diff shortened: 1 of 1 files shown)


org/Hibachi/HibachiSmartList.cfc
```diff
@@ -810,6 +810,14 @@ component accessors="true" persistent="false" output="false" extends="HibachiObj
 		return variables.pageRecords;
 	}
 	
+	public any function getFirstRecord(boolean refresh=false) {
+		if( !structKeyExists(variables, "firstRecord") || arguments.refresh == true) {
```
pjohn1024: I think we should just write arguments.refresh instead of checking for arguments.refresh == true

PR #4845, opened 2016-09-22
Split Shipping Functionality in Shipping Service & International USPS

jasperboyd:
This change is [Reviewable](https://reviewable.io/reviews/ten24/slatwall/4845)


integrationServices/usps/IntlRatesV2RequestTemplate.cfm
```diff
@@ -0,0 +1,84 @@
...
+	provided that you include the source code of that other code when and as the 
+	GNU GPL requires distribution of source code.
+    
+    If you modify this program, you may extend this exception to your version 
+    of the program, but you are not obligated to do so.
+
+	Notes:
+	
+--->
+<cfsilent>
```
rmarchandTen24: i think it is unnecessary to maintain a completely separate template for international when it has such similar content to the RateV4Response template.

integrationServices/usps/Shipping.cfc
```diff
@@ -70,10 +70,59 @@ component accessors="true" output="false" displayname="USPS" implements="Slatwal
 			30 = "Express Mail - Legal Flat Rate Evelope",
 			31 = "Express Mail - Legal Flat Rate Envelope, Hold For Pickup",
 			32 = "Express Mail - Legal Flat Rate Envelope, Sunday/Holiday Delivery",
+			33 = "Priority Mail - Hold For Pickup",
```
rmarchandTen24: this may not be possible but is there a way to get these options from usps rather than hard code them?

integrationServices/usps/Shipping.cfc
```diff
@@ -96,18 +145,22 @@ component accessors="true" output="false" displayname="USPS" implements="Slatwal
         	requestURL = setting("liveAPIEndPointURL");
         }        
         
+        var xmlPacket = "";
+		var response = ""; 
         if(arguments.requestBean.getShipToCountryCode() == "US") {
```
rmarchandTen24: you can see that US is used to drive the state of whether this is an international rate. this could be a state on the object.

model/service/ShippingService.cfc
```diff
@@ -74,10 +112,56 @@ component extends="HibachiService" persistent="false" accessors="true" output="f
 			var ratesRequestBean = getTransient("ShippingRatesRequestBean");
 			ratesRequestBean.populateWithOrderFulfillment(arguments.orderFulfillment);
 
+			var splitShipmentFlag = false;
```
rmarchandTen24: ideally splitshipmentflag is a state instead of a variable of the process. This makes the code more procedural and less reusable as this expands

integrationServices/usps/IntlRatesV2RequestTemplate.cfm
jasperboyd: It's actually is a completely different API even though they share a lot of the same traits, but there are some that Domestic doesn't have, when I was handed off this I believe @sumitverma  said to make a separate template

integrationServices/usps/Shipping.cfc
jasperboyd: I don't see anything on the api listing page https://www.usps.com/business/web-tools-apis/documentation-updates.htm

integrationServices/usps/Shipping.cfc
jasperboyd: Yeah that makes sense to me I'll make this change

model/service/ShippingService.cfc
jasperboyd: This one I'm not so sure about, I see what you're saying but the state has to be derived from a combination of the shippingMethodRate, and the order fulfillment Items theres no object where it can be stored. I'd also say the entire shipping service is pretty much procedural already

PR #4853, opened 2016-09-28
Develop amazon session updates

ian-a-hickey:
This pull addresses Greg remaining concerns by removing the unneeded getLoggedInFlag method in hibachiScope and by adding a non-persistent property for clients to use. There is still one more issue Greg wants to address but as its due to duplicate code rather than anything functional its not needed immediately.

---

This change is [Reviewable](https://reviewable.io/reviews/ten24/slatwall/4853)


gregmoser:
@ian-a-hickey let me know when you're ready to go over this PR, lets talk before it gets merged in because I still want to see that other behavior at some point.

PR #4857, opened 2016-09-28
adding form email confirmation value to be the target for recipient

jasperboyd:
This change is [Reviewable](https://reviewable.io/reviews/ten24/slatwall/4857)


model/entity/FormResponse.cfc
```diff
@@ -79,6 +79,15 @@ component displayname="FormResponse" entityname="SlatwallFormResponse" table="Sw
         return variables.attributes;
 	} 
 
+	public any function getFormEmailConfirmationValue(){
+ 		var attributeValueSmartList = getPropertySmartList('attributeValues'); 
+ 		attributeValueSmartList.addFilter('attribute.formEmailConfirmationFlag', true); 
+ 		var records = attributeValueSmartList.getRecords();
```
rmarchandTen24: if you are only interested in one record hibachismartlist now has a function called getFirstRecord.
it should return the record or a null

jasperboyd: nice! I wasn't a fan of the way this was written

PR #4900, opened 2016-10-12
Update HibachiService

ian-a-hickey:
This fixes an issue in a custom project where the listing display (using attributes/custom properties) was throwing an error on page load on every page that  had the display.

---
This change is [Reviewable](https://reviewable.io/reviews/ten24/slatwall/4900)


model/service/HibachiService.cfc
```diff
@@ -147,8 +147,10 @@ component accessors="true" output="false" extends="Slatwall.org.Hibachi.HibachiS
 	public any function getPropertiesWithAttributesByEntityName(required string entityName){
 		var entityObject = getEntityObject( arguments.entityName );
 		var properties = entityObject.getFilterProperties();
-		
-		return properties;
+		if (!isNull(properties)){
```
rmarchandTen24: filter properties are an array so null check shouldn't apply

PR #4922, opened 2016-10-19
OrderService hotfix

pauaware:
add conditions for where shipping method is null (having not been set yet)

---

This change is [Reviewable](https://reviewable.io/reviews/ten24/slatwall/4922)


rmarchandTen24:
so there are cases in updateOrderFulfillmentShippingMethodOptions method where it does work with shippingmethod as null. When/how did you experience your error?

PR #4927, opened 2016-10-19
Added validation to prevent location addresses from being deleted when

christopher-kundrat:
associated to an order. Created a function in location address to handle
the logic and an rbkey to display the error message

---
This change is [Reviewable](https://reviewable.io/reviews/ten24/slatwall/4927)


model/entity/LocationAddress.cfc
```diff
@@ -106,6 +106,21 @@ component displayname="Location Address" entityname="SlatwallLocationAddress" ta
 	}
 	
 	// =============  END:  Bidirectional Helper Methods ===================
+	
+	// =============== START: Custom Validation Methods ====================
+	public boolean function isDeletable() {
```
rmarchandTen24: https://github.com/ten24/slatwall/blob/master/org/Hibachi/HibachiEntity.cfc#L233

this will override validation entirely

PR #4929, opened 2016-10-20
Created new non-persistent property to get the first scheduled sku of a

christopher-kundrat:
product. Also added a delete override to make sure that if the last sku
is being deleted then we need to delete the product schedule as well

---
This change is [Reviewable](https://reviewable.io/reviews/ten24/slatwall/4929)

### DIFF
--- a/model/service/SkuService.cfc
+++ b/model/service/SkuService.cfc
@@ -586,6 +586,27 @@ component extends="HibachiService" persistent="false" accessors="true" output="f
 	}
 
 	// ======================  END: Save Overrides ============================
+	
+	// ====================== START: Delete Overrides =========================
+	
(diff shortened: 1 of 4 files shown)


model/service/SkuService.cfc
```diff
@@ -586,6 +586,27 @@ component extends="HibachiService" persistent="false" accessors="true" output="f
 	}
 
 	// ======================  END: Save Overrides ============================
+	
+	// ====================== START: Delete Overrides =========================
+	
+	public boolean function deleteSku(required any sku) {
+
+		// Check delete validation
+		if(arguments.sku.isDeletable() && !isNull(arguments.sku.getProductSchedule())) {
```
rmarchandTen24: this should be a condition in the validation json

PR #4931, opened 2016-10-20
Hotfix ck disable create event

christopher-kundrat:
This change is [Reviewable](https://reviewable.io/reviews/ten24/slatwall/4931)


rmarchandTen24:
shouldn't this be a backend validation?

christopher-kundrat:
@rmarchandTen24  I believe it's already a backend validation. It's the buttons on the product listing page that needs to have an inline disable

PR #4951, opened 2016-10-25
Hotfix ih subscription usage amount bug

ian-a-hickey:
This fixes the issue where the renewal price (and renewal sku price) was never being used on a renewal order.

---

This change is [Reviewable](https://reviewable.io/reviews/ten24/slatwall/4951)

### DIFF
--- a/model/entity/SubscriptionUsage.cfc
+++ b/model/entity/SubscriptionUsage.cfc
@@ -180,8 +180,15 @@ component entityname="SlatwallSubscriptionUsage" table="SwSubsUsage" persistent=
 	}
 
 	public numeric function getRenewalPrice(){
+		if (structKeyExists(variables, "renewalPrice")){
+			return variables.renewalPrice;
+		}
 		if(!structKeyExists(variables, "renewalPrice") && !isNull(this.getRenewalSku())){
-			variables.renewalPrice = this.getRenewalSku().getRenewalPrice();
+			if (this.getRenewalSku().getRenewalPrice() > 0){
+				variables.renewalPrice = this.getRenewalSku().getRenewalPrice();
+			}else{
+				variables.renewalPrice = this.getRenewalSku().getPrice();
+			}
 		}else{
 			variables.renewalPrice = 0;
 		}
--- a/model/entity/SubscriptionUsage.cfc
+++ b/model/entity/SubscriptionUsage.cfc
@@ -144,8 +144,8 @@ component entityname="SlatwallSubscriptionUsage" table="SwSubsUsage" persistent=
 		
 		//if we have a renewal sku, then use that to get the renewal price (still by currency code)
 		if (!isNull(arguments.orderItem.getSku().getRenewalSku())){
-			subscriptionUsage.setRenewalSku(arguments.orderItem.getSku().getRenewalSku());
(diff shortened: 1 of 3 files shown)


model/entity/SubscriptionUsage.cfc
```diff
@@ -180,8 +180,15 @@ component entityname="SlatwallSubscriptionUsage" table="SwSubsUsage" persistent=
 	}
 
 	public numeric function getRenewalPrice(){
+		if (structKeyExists(variables, "renewalPrice")){
+			return variables.renewalPrice;
+		}
 		if(!structKeyExists(variables, "renewalPrice") && !isNull(this.getRenewalSku())){
```
sumitverma: I think this should be else if

model/entity/SubscriptionUsage.cfc
```diff
@@ -180,8 +180,15 @@ component entityname="SlatwallSubscriptionUsage" table="SwSubsUsage" persistent=
...
-			variables.renewalPrice = this.getRenewalSku().getRenewalPrice();
+			if (this.getRenewalSku().getRenewalPrice() > 0){
+				variables.renewalPrice = this.getRenewalSku().getRenewalPrice();
+			}else{
+				variables.renewalPrice = this.getRenewalSku().getPrice();
+			}
```
sumitverma: I don't believe we need this check. If renewal price is set to 0 then that's what we need to use, we can't just use price

ian-a-hickey: Renewal Price is always set to zero as the default. Because of that, there is no way to tell if the user intends to use zero for the renewal price when specifying a sku. So instead, I just use the price, which can be zero if thats what you intend it to be. Does that make sense?

model/entity/SubscriptionUsage.cfc
```diff
@@ -180,8 +188,13 @@ component entityname="SlatwallSubscriptionUsage" table="SwSubsUsage" persistent=
 	}
 
 	public numeric function getRenewalPrice(){
-		if(!structKeyExists(variables, "renewalPrice") && !isNull(this.getRenewalSku())){
+		//if we have a renewal price, then use it.
+		if (structKeyExists(variables, "renewalPrice")){
+			return variables.renewalPrice;
+		//otherwise check if we have a renewal sku.
+		}else if(!structKeyExists(variables, "renewalPrice") && !isNull(this.getRenewalSku())){
```
ian-a-hickey: I added an else if here.

model/service/SubscriptionService.cfc
```diff
@@ -141,20 +141,19 @@ component extends="HibachiService" persistent="false" accessors="true" output="f
 		subscriptionUsage.setExpirationDate( arguments.orderItem.getSku().getSubscriptionTerm().getInitialTerm().getEndDate() );
 		subscriptionUsage.setNextBillDate( subscriptionUsage.getExpirationDate() );
 		subscriptionUsage.setFirstReminderEmailDateBasedOnNextBillDate();
-		subscriptionUsage.setRenewalSku(arguments.orderItem.getSku().getRenewalSku());
```
ian-a-hickey: this was moved to be handled inside the copyFrom in subscriptionUsage

ian-a-hickey: Tested that this is still working with the three changes.

PR #4961, opened 2016-10-26
Update orderFulfillment

ian-a-hickey:
cascade delete here causes other related issues so I'm removing it.

---

This change is [Reviewable](https://reviewable.io/reviews/ten24/slatwall/4961)


gregmoser:
This has the potential now to leave orphaned records no?  should it be "all-delete-orphan"?

sumitverma:
No, there is another one-to-many for all shipping method options that will take care of deletes. This property is just a many-to-one for the selected one.

gregmoser:
Gotcha, thanks @sumitverma

PR #4962, opened 2016-10-26
adding model definition for AccountCollection and relations with Account and Collection

jasperboyd:
This change is [Reviewable](https://reviewable.io/reviews/ten24/slatwall/4962)


model/entity/AccountCollection.cfc
```diff
@@ -0,0 +1,118 @@
...
+	property name="remoteID" ormtype="string";
+
+	// Audit Properties
+	property name="createdDateTime" hb_populateEnabled="false" ormtype="timestamp";
+	property name="createdByAccountID" hb_populateEnabled="false" ormtype="string";
+	property name="modifiedDateTime" hb_populateEnabled="false" ormtype="timestamp";
+	property name="modifiedByAccountID" hb_populateEnabled="false" ormtype="string";
+
+	// Non-Persistent Properties
+	property name="collectionConfigStruct" type="struct" persistent="false";
```
jasperboyd: Makes sense to me, I just brought these over from Collection.cfc for convenience

model/process/AccountCollection_Create.cfc
```diff
@@ -0,0 +1,68 @@
...
+	property name="account";
+	property name="collection";
+
+
+	public any function getAccountCollection(){
+		if(!isNull(this.getAccount())){
+			variables.accountCollection.setAccount(this.getAccount());
+		}
+		if(!isNull(this.getCollection())){
+			variables.accountCollection.setCollection(this.getCollection());
```
rmarchandTen24: I think that this method should check if the collection is new or not. If getNewFlag then copy over the collectionConfig if persisted then setCollection

model/validation/AccountCollection.json
```diff
@@ -0,0 +1,8 @@
+{
+	"properties" : {
+		"collectionDisplayCode":	[{"contexts":"save","uniqueOrNull":true,"regex":"^[a-zA-Z0-9-_.|:~^]+$"}],
+		"collectionConfig":			[{"contexts":"save","required":true}],
```
rmarchandTen24: collectionConifg is not a required field. It will either be save a collection or a config

jasperboyd: @rmarchandTen24 the spec says it is required

sumitverma: it's conditionally required. If there is no collection specified then collectionConfig is required.

jasperboyd: okay I'll update the spec

rmarchandTen24: k let me know when you update the validation

org/Hibachi/HibachiCollectionService.cfc
```diff
@@ -377,35 +377,35 @@ component output="false" accessors="true" extends="HibachiService" {
 			baseEntityAlias = '_' & lcase(rc.entityName)
 		};
 
-		if(!isnull(arguments.rc.filterConfig)){
+		if(!isnull(arguments.data.filterConfig)){
```
rmarchandTen24: Change makes sense, but there are still references to rc in the function and the rc is still part of the arguments scope. Just change that to data and make sure all calls to this function don't explicitly say (rc=rc)

rmarchandTen24: ideally make a unit test for this function to make sure we can catch this as a regression in the future

PR #4963, opened 2016-10-26
able to removefilter from collection

rmarchandTen24:
This change is [Reviewable](https://reviewable.io/reviews/ten24/slatwall/4963)

### DIFF
--- a/model/entity/Collection.cfc
+++ b/model/entity/Collection.cfc
@@ -100,12 +100,45 @@ component displayname="Collection" entityname="SlatwallCollection" table="SwColl
 	property name="collectionObjectOptions" persistent="false";
 
 	// ============ START: Non-Persistent Property Methods =================
+	
+	public any function init(){
+		super.init();
+
+		param name="session.entityCollection" type="struct" default="#structNew()#";
+		param name="session.entityCollection.savedStates" type="array" default="#arrayNew(1)#";
+
+		variables.hqlParams = {};
+		variables.hqlAliases = {};
+		variables.Cacheable = false;
+		variables.CacheName = "";
+		variables.currentPageDeclaration = 1;
+		variables.pageRecordsStart = 1;
+		variables.pageRecordsShow = 10;
+		variables.keywords = "";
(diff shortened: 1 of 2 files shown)


model/entity/Collection.cfc
```diff
@@ -100,13 +100,46 @@ component displayname="Collection" entityname="SlatwallCollection" table="SwColl
 	property name="collectionObjectOptions" persistent="false";
 
 	// ============ START: Non-Persistent Property Methods =================
+	
+	public any function init(){
+		super.init();
+
+		param name="session.entityCollection" type="struct" default="#structNew()#";
+		param name="session.entityCollection.savedStates" type="array" default="#arrayNew(1)#";
```
sumitverma: We shouldn't be using direct reference to session. anywhere. We can use getHibachiScope().setSessionValue() if needed. Infact we want to avoid using session all together because it doesn't scale well on multi server instance.

rmarchandTen24:
agreed, I didn't add this but moved the init function to the top. This may have
been added previously without being fully implemented
 





On Wed, Oct 26, 2016 1:00 PM, Sumit Verma notifications@github.com
wrote:
@sumitverma  commented on this pull request.


--------------------------------------------------------------------------------

In model/entity/Collection.cfc:

> @@ -100,13 +100,46 @@ component displayname="Collection" entityname="SlatwallCollection" table="SwColl
 	property name="collectionObjectOptions" persistent="false";
 
 	// ============ START: Non-Persistent Property Methods =================
+	
+	public any function init(){
+		super.init();
+
+		param name="session.entityCollection" type="struct" default="#structNew()#";
+		param name="session.entityCollection.savedStates" type="array" default="#arrayNew(1)#";


We shouldn't be using direct reference to session. anywhere. We can use
getHibachiScope().setSessionValue() if needed. Infact we want to avoid using
session all together because it doesn't scale well on multi server instance.

sumitverma:
Ya, I think it's a carry over from smartlist. I will add a task.

Sumit Verma

PR #4964, opened 2016-10-26
updates to fix issue deleting orderitems.

ian-a-hickey:
Because orderfulfillment has a calculated property that is a relationship rather than a simple property (to selectedShippingMethodOption), entitymerge needed to be added to flushORMSession to get the entity back onto the session after the initial flush.

Once that was fixed, a constraint was being thrown from orderFulfillment to shippingMethodOptionID. I fixed that by (after trying javacast null on orderFulfillment side and the other way around) adding a DAO method to remove the id from orderFulfillment. With that working, removing orderItems works again.

Thanks for the help Sumit.

---

This change is [Reviewable](https://reviewable.io/reviews/ten24/slatwall/4964)

### DIFF
--- a/model/entity/Order.cfc
+++ b/model/entity/Order.cfc
@@ -1116,6 +1116,11 @@ component displayname="Order" entityname="SlatwallOrder" table="SwOrder" persist
 		arguments.orderFulfillment.setOrder( this );
 	}
 	public void function removeOrderFulfillment(required any orderFulfillment) {
+		//must do this to handle the related calculated property on orderFulfillment.
+		if (!isNull(arguments.orderFulfillment.getSelectedShippingMethodOption())){
+			var dao = getDao("OrderDAO");
+			dao.removeSelectedShippingMethodOptionFromOrderFulfillments(arguments.orderFulfillment);
+		}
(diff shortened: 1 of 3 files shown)


model/entity/Order.cfc
```diff
@@ -1116,6 +1116,11 @@ component displayname="Order" entityname="SlatwallOrder" table="SwOrder" persist
 		arguments.orderFulfillment.setOrder( this );
 	}
 	public void function removeOrderFulfillment(required any orderFulfillment) {
+		//must do this to handle the related calculated property on orderFulfillment.
+		if (!isNull(arguments.orderFulfillment.getSelectedShippingMethodOption())){
+			var dao = getDao("OrderDAO");
+			dao.removeSelectedShippingMethodOptionFromOrderFulfillments(arguments.orderFulfillment);
+		}
```
sumitverma: typically we don't overide the remove method for this kind of behavior. I think this belongs in a service delete overide method.  @rmarchandTen24 please review.

rmarchandTen24: we usually don't. I think that this may be because of the whole calculated entity property. Might need to set the nonpersistent to null as well

PR #4972, opened 2016-10-27
Hotfix ih subscription usage

ian-a-hickey:
I had to update both the subscription renew process object and the service in order to get subscription renewals working on the frontend. The reason was because A. a new order object was not being created and B. the service would never call place order (logic just skipping in this case). This is how I was able to get it working for testing, but I don't think its a great solution and want to discuss. So, I'm putting this here for review.

---

This change is [Reviewable](https://reviewable.io/reviews/ten24/slatwall/4972)

### DIFF
--- a/model/service/SubscriptionService.cfc
+++ b/model/service/SubscriptionService.cfc
@@ -523,16 +523,21 @@ component extends="HibachiService" persistent="false" accessors="true" output="f
 						orderPayment.copyFromOrderPayment( arguments.processObject.getOrderPayment() );
 						orderPayment.setCurrencyCode( order.getCurrencyCode() );
 						orderPayment.setOrder( order );
-					} else if (arguments.processObject.getRenewalPaymentType() eq 'new') {
+					}  else if (arguments.processObject.getRenewalPaymentType() eq 'new') {
+						
 						order = getOrderService().processOrder(order, arguments.data, 'addOrderPayment');
+						getOrderService().saveOrder(order);
 					}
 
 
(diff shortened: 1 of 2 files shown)


model/service/SubscriptionService.cfc
```diff
@@ -523,16 +523,21 @@ component extends="HibachiService" persistent="false" accessors="true" output="f
 						orderPayment.copyFromOrderPayment( arguments.processObject.getOrderPayment() );
 						orderPayment.setCurrencyCode( order.getCurrencyCode() );
 						orderPayment.setOrder( order );
-					} else if (arguments.processObject.getRenewalPaymentType() eq 'new') {
+					}  else if (arguments.processObject.getRenewalPaymentType() eq 'new') {
+						
 						order = getOrderService().processOrder(order, arguments.data, 'addOrderPayment');
+						getOrderService().saveOrder(order);
```
sumitverma: Why do we need to save the order here?

model/service/SubscriptionService.cfc
```diff
@@ -523,16 +523,21 @@ component extends="HibachiService" persistent="false" accessors="true" output="f
...
 					}
 
 
-					if(!isNull(orderPayment)){
+					if(!isNull(orderPayment) || (!isNull(order) && !isNull(order.getOrderPayments()[1]))){
```
sumitverma: Where is the orderPayment coming from here? Seems like it will always be null. Also, we can't just get the first one from order, it needs to be the first active one.

PR #4977, opened 2016-10-28
Puerto rico is considered to be a US rate request by USPS

jasperboyd:
This change is [Reviewable](https://reviewable.io/reviews/ten24/slatwall/4977)


sumitverma:
@jasperboyd @rmarchandTen24 this change in not right. ShippingRequestBean is used by all integrations and PR may not be considered international by all. At the very least this needs to be overridden in usps integration.

PR #5002, opened 2016-11-10
Updates HibachiService

ian-a-hickey:
This is the version of the export function we use in a custom project.

---
This change is [Reviewable](https://reviewable.io/reviews/ten24/slatwall/5002)


sumitverma:
@ian-a-hickey What's this change for? I know export works fine currently.

PR #5005, opened 2016-11-11
Fixing an issue where multiple associations were being created on swcontentcategory

dzikakis:
Won't clean up any existing repeated rows in the DB though.. I imagine you'll want to do a quick update script for that: Truncate swcontentcategory and let it rebuild.

---
This change is [Reviewable](https://reviewable.io/reviews/ten24/slatwall/5005)

### DIFF
--- a/integrationServices/mura/model/handler/MuraEventHandler.cfc
+++ b/integrationServices/mura/model/handler/MuraEventHandler.cfc
@@ -1271,6 +1271,8 @@
 					count(SwContent.contentID) as missingCount
 				FROM
 					tcontentcategoryassign
+				  INNER JOIN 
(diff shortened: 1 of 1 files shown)


gregmoser:
Actually, now I'm confused by this.  If the assigned category relationship exists on the mura side, regardless of if the content is active... shouldn't it be there on the Slatwall side?

Doesn't seem like this would have anything to do with nested auto-assignment of categories on the Slatwall side.

dzikakis:
Greg, what I had said over hangouts; that the subcategories were being
mapped on the slatwall side, wasn't entirely accurate it turns out. The
mura content I was looking at had been assigned the subcategories in the
past. The issue is that the sync pulls all the historical versions of the
mura content, and the historical categories.

gregmoser:
I see, ok.  And the "active" value is set to 1 on the only piece of content that is active... got it.

PR #5018, opened 2016-11-18
updating address simple rep to a calculated field

rmarchandTen24:
This change is [Reviewable](https://reviewable.io/reviews/ten24/slatwall/5018)


sumitverma:
Some projects could have 100K+ address records. The migration script will probably need to get optimized. May be override get simplerepresentation to use the addressName method if the calculated property is null?

PR #5042, opened 2016-12-03
fix for null active flag

sumitverma:
This change is [Reviewable](https://reviewable.io/reviews/ten24/slatwall/5042)


gregmoser:
@sumitverma we should be using the activeFlags database script that automatically sets defaults.  I have a feeling that some of the new activeFlag properties were never added to that file.

sumitverma:
This is a side effect of data import. Good to do null check anyways


Sumit Verma

gregmoser:
Data import will be caught by the activeFlag script anyway because it's run on reload... yes it is good to do the null check anyway, but we should still set defaults for these active flags in the script

PR #5049, opened 2016-12-07
Collection Improvements including Dirty Read, Exclude Order/Group by for Aggregates

jasperboyd:
This change is [Reviewable](https://reviewable.io/reviews/ten24/slatwall/5049)


model/entity/Collection.cfc
```diff
@@ -428,11 +434,7 @@ component displayname="Collection" entityname="SlatwallCollection" table="SwColl
 		}
 
 		var propertyIdentifier = listFirst(arguments.orderByString,'|');
-		var direction = 'ASC';
```
rmarchandTen24: this change would make the code expect a pipe no matter what. The previous code defaulted to ASC if the pipe wasn't supplied.

model/entity/Collection.cfc
```diff
@@ -986,11 +986,11 @@ component displayname="Collection" entityname="SlatwallCollection" table="SwColl
 		return fromHQL;
 	}
 
-	public string function getHQL(boolean excludeSelectAndOrderBy = false, forExport=false){
+	public string function getHQL(boolean excludeSelectAndOrderBy = false, forExport=false, removeOrderBy = false){
```
rmarchandTen24: we have a flag for excludeselectandorderby , shouldn't this do the work of removeOrderBy?

model/entity/Collection.cfc
```diff
@@ -1156,6 +1157,9 @@ component displayname="Collection" entityname="SlatwallCollection" table="SwColl
 	}
 
 	private string function getOrderByHQL(array orderBy=[]){
+		if(structKeyExists(variables, 'orderByRequired') && !variables.orderByRequired){
```
rmarchandTen24: just curious what conditions an order by is required?

model/entity/Collection.cfc
jasperboyd: That makes sense to me, I believe this was @migueltarga 's change this pull request has both our code, he may have a reason for it

model/entity/Collection.cfc
jasperboyd: You'll see below beginning on line 1660 if the request is only min / max aggregate functions the order by and group by will cause the results to be incorrect and instead of turning only one record it will turn a page record for each and every result with incorrect aggregate values.

model/entity/Collection.cfc
jasperboyd: This was from an earlier implementation I had done that we found not to handle all cases it needed too, I will remove these lingering references to it

model/entity/Collection.cfc
migueltarga: Its because we are in version `4.5.020.037`

https://github.com/ten24/slatwall/commit/67a89e6d09cece81ec7ae039a814be05c078bfe3

jasperboyd: got it I'll add it back in thanks @migueltarga

model/entity/Collection.cfc
jasperboyd: The reason the order by is necessary is because otherwise we always tack on order by createdDateTime.

Ex:
`select min(price) as minPrice from SwSku order by createdDateTime`

Throws 

> Column "SwSku.createdDateTime" is invalid in the ORDER BY clause because it is not contained in either an aggregate function or the GROUP BY clause.

We can't have a group by because it skews the results to be individual pageRecords instead of a single aggregate. 

Not exactly sure how to convey that with a unit test

PR #5053, opened 2016-12-07
adding improvements to tabs so that you can hide them...

jasperboyd:
and select only visible tabs as active when initializing

---
This change is [Reviewable](https://reviewable.io/reviews/ten24/slatwall/5053)


gregmoser:
This seems cool.

PR #5061, opened 2016-12-12
Preventing non-numeric input on the stock adjustment modal fixing bug with subsequent save requests

jasperboyd:
This change is [Reviewable](https://reviewable.io/reviews/ten24/slatwall/5061)


org/Hibachi/client/src/core/services/hibachiservice.ts
```diff
@@ -142,6 +142,8 @@ class HibachiService{
 	populateEntity = (entityName, data)=>{ 
 		var newEntity = this.newEntity(entityName); 
 		angular.extend(newEntity.data,data); 
+		//make sure all levels are populated
+		angular.extend(newEntity,data);
```
rmarchandTen24: what data would this be used for? example?

PR #5065, opened 2016-12-12
ensuring that calculated properties update with proper values when settings change

jasperboyd:
This change is [Reviewable](https://reviewable.io/reviews/ten24/slatwall/5065)


rmarchandTen24:
👍

rmarchandTen24:
@jasper @KevinJason Can you create a new branch of hotfix without this commit https://github.com/ten24/slatwall/pull/5065/commits/de23ad2c2a5d7cc4007691ac47e8acf0f16db1b4

this will cause those items to be removed from develop when hotfix is merged into develop. I've actually already seen this manifest before where settings and rbkeys are missing in develop after a merge.

PR #5083, opened 2016-12-19
Hotfix rm luceebugextendedprice

rmarchandTen24:
This change is [Reviewable](https://reviewable.io/reviews/ten24/slatwall/5083)


gregmoser:
@rmarchandTen24 we need to do this on EVERY precisionEvaluate in the application.  @sumitverma can you confirm that is what we need to do?

ten24git:
Correct. We will need to fix all. Probably will make sense to do that in
develop and not hotfix.

Sumit Verma

PR #5085, opened 2016-12-19
fixing blank fields when adding products to content 

jasperboyd:
This change is [Reviewable](https://reviewable.io/reviews/ten24/slatwall/5085)

### DIFF
--- a/admin/client/src/content/components/assignedproducts.html
+++ b/admin/client/src/content/components/assignedproducts.html
@@ -8,9 +8,7 @@
                     data-show-order-by="false"
                     data-administrative-count="1"
                     data-sortable-field-name="productListingPageSortOrder"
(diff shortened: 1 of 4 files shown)


admin/client/src/content/components/assignedproducts.html
```diff
@@ -30,7 +32,7 @@
                                                       data-is-visible="true">
                                 </sw-collection-column>
                                 <sw-collection-column data-property-identifier="product.productType.productTypeName"  
-                                                      data-fallback-property-identifiers="product.productType.productTypeName" 
+                                                      data-fallback-property-identifiers="productType_productTypeName"
```
rmarchandTen24: ??? doesn't this break convention by using the alias vs a propertyIdentifier?

PR #5086, opened 2016-12-20
estimated shipping

ian-a-hickey:
This pull needed to be against hotfix as well as develop so that I may pull it into a project.

---
This change is [Reviewable](https://reviewable.io/reviews/ten24/slatwall/5086)

### DIFF
--- a/model/service/PublicService.cfc
+++ b/model/service/PublicService.cfc
@@ -1164,4 +1164,72 @@ component extends="HibachiService"  accessors="true" output="false"
         arguments.data.ajaxResponse['countryCodeOptions'] = getAddressService().getCountryCodeOptions();
     }
     
+    /** Given a skuCode and/or address information - returns the estimated shipping rates for that sku. */
+    public any function getEstimatedShippingCostBySkuCode(any data){
+    	if (!isNull(data.skuCode)){
+    		
+    		//data setup.
+    		var orderFulfillment = getService("OrderService").newOrderFulfillment();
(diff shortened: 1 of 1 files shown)


model/service/PublicService.cfc
```diff
@@ -1164,4 +1164,72 @@ component extends="HibachiService"  accessors="true" output="false"
         arguments.data.ajaxResponse['countryCodeOptions'] = getAddressService().getCountryCodeOptions();
     }
     
+    /** Given a skuCode and/or address information - returns the estimated shipping rates for that sku. */
+    public any function getEstimatedShippingCostBySkuCode(any data){
+    	if (!isNull(data.skuCode)){
```
rmarchandTen24: when not dealing with an object property. I suggest using structKeyExists over !isNull

model/service/PublicService.cfc
```diff
@@ -1164,4 +1164,72 @@ component extends="HibachiService"  accessors="true" output="false"
...
+    						if (structKeyExists(rate, "shippingMEthodCode")){
+    							options['#rate.shippingMethodCode#'] = rate;
+    						}
+    					}
+    				}
+    			}
+    		}
+    		
+    		//remove the orderfulfillment that we used to get the rates because it will disrupt other entities saving.
+    		getService("OrderService").deleteOrderFulfillment(orderFulfillment);
```
rmarchandTen24: we should avoid having to create an order fulfillment to then delete it. Are you able to break apart the function so that  this is not required? Maybe going through a service method?

PR #5101, opened 2016-12-22
Auto generate aliases and add joins on Collections addOrderBy

migueltarga:
This change is [Reviewable](https://reviewable.io/reviews/ten24/slatwall/5101)


meta/tests/unit/entity/CollectionTest.cfc
```diff
@@ -435,6 +435,11 @@ component extends="Slatwall.meta.tests.unit.entity.SlatwallEntityTestBase" {
 		myCollection.addFilter('productDescription',uniqueNumberForDescription);
 		myCollection.setOrderBy('productName|asc');
 
+
+		request.debug(myCollection.getCollectionConfigStruct());
```
rmarchandTen24: no biggie but just a heads up. We need to remove the request.debug because the headless unit test runner will error with wtf is request.debug.

PR #5158, opened 2017-01-10
wrapped all precisionEvaluate calls in val() except calls internal to…

gerickson808:
… another precisionEvaluate

---
This change is [Reviewable](https://reviewable.io/reviews/ten24/slatwall/5158)

### DIFF
--- a/admin/views/entity/preprocessorder_createreturn.cfm
+++ b/admin/views/entity/preprocessorder_createreturn.cfm
@@ -95,7 +95,7 @@ Notes:
 							<td>#orderItem.getSku().getSkuDefinition()#</td>
 							<td>#orderItem.getQuantity()#</td>
 							<td>#orderItem.getQuantityDelivered()#</td>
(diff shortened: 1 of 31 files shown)


rmarchandTen24:
@gregmoser The nuance is a bit lost on me? What do the quotes do here?

val(precisionEvaluate("round(price * val(getQuantity()) * 100) / 100"));

sumitverma:
It was based on lot of testing we had done and the fact that
precisionevaluate expects a string expression. But, we can skip adding
quotes, because based on documentation it is more performant to not add
quotes.

https://helpx.adobe.com/coldfusion/cfml-reference/coldfusion-functions/functions-m-r/precisionevaluate.html

dzikakis:
FYI: The question of quotes brought back some recollection of this: https://groups.google.com/forum/#!searchin/slatwallecommerce/precisionEvaluate%7Csort:relevance/slatwallecommerce/t3znQ7XcDAI/ydD-JWcgKQQJ 

Seems it was a pre-4.2 Railo issue

gregmoser:
@rmarchandTen24 and @sumitverma the quotes are important because without them it is inconsistent between Adobe & Lucee as not both engines perform the same style of type check optimization.

PR #5163, opened 2017-01-11
adding cancel functionality to cart controller

jasperboyd:
This change is [Reviewable](https://reviewable.io/reviews/ten24/slatwall/5163)


public/controllers/cart.cfc
```diff
@@ -155,6 +155,15 @@ component output="false" accessors="true" extends="Slatwall.org.Hibachi.HibachiC
   		getService("PublicService").addOrderItems(rc);
   		
   	}
+
+	public void function cancel(required any rc) {
```
sumitverma: I would recommend changing to cancelOrder

jasperboyd: yeah I agree that's better

PR #5169, opened 2017-01-12
adds a lock to addOrderItem to fix issue where cart get overridden if

ian-a-hickey:
users add items to the cart too fast. This happens when we have a design
for a product page that has multiple add order item buttons. If the user
adds one item and then quickly adds another item, the final cart results
in only the last item added instead of all of the items. Adding the lock
fixes the problem as the same users request will wait until the previous
one is finished.
Updates to add functionality to the addOrderITems method in publicservice. This allows the user to pass the skuIds or skuCodes with or without a list of quantities.
slataction=public:cart.addOrderItems&skuIds=402828ee57e7a75b0157f86079ee0319,402828ee57e7a75b0157f860c4b7031f&quantities=1,2
or
?slataction=public:cart.addOrderItems&skuCodes=someCode1,someCode2&quantities=1,2
or
?slataction=public:cart.addOrderItems&skuIds=402828ee57e7a75b0157f86079ee0319,402828ee57e7a75b0157f860c4b7031f which assumes a quantity of 1 for each item.

This change is [Reviewable](https://reviewable.io/reviews/ten24/slatwall/5169)


gregmoser:
This fix does not work in a load balanced environment as locks are only tied to 1 server.  Also, just an FYI... @sumitverma hates locks ;)

I believe the the crux of this issue, is that there are 2 "new" carts at the same time because this first button click & second button click both are being interpreted as the "first" item being added to a cart.

In fact this is a bigger issue, because the session is potentially brand new on both of those requests unless the "Ignore Session Tracking until Logged In or Cart Created" is set to false as a global setting.

@sumitverma, @rmarchandTen24 and myself should talk about the best solution to this issue with you.

ian-a-hickey:
Updated the pull request to update addOrderItems to accept a list of quantities instead of a lock on addOrderItem.

dzikakis:
This might be considered a cry from the peanut gallery, but any chance addOrderItems could be built to handle stocks or skus with location?

PR #5201, opened 2017-01-23
Hotfix ih dupe shipping rates

ian-a-hickey:
Issue: While creating the array of shipping method options to return, it is adding multiple of the same option.

Fix: I fixed this by adding a method that checks for that option in an option array, and then used that method three places.

---
This change is [Reviewable](https://reviewable.io/reviews/ten24/slatwall/5201)

### DIFF
--- a/model/entity/OrderFulfillment.cfc
+++ b/model/entity/OrderFulfillment.cfc
@@ -375,10 +375,10 @@ component displayname="Order Fulfillment" entityname="SlatwallOrderFulfillment" 
 
     			for(var i=1; i<=arrayLen(optionsArray); i++) {
     				var thisExistingOption = optionsArray[i];
-
-    				if( ((sortType eq 'price' && thisOption.totalCharge < thisExistingOption.totalCharge)
-    				  	||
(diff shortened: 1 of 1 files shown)


model/entity/OrderFulfillment.cfc
```diff
@@ -376,17 +376,19 @@ component displayname="Order Fulfillment" entityname="SlatwallOrderFulfillment"
     			for(var i=1; i<=arrayLen(optionsArray); i++) {
     				var thisExistingOption = optionsArray[i];
 
-    				if( (sortType eq 'price' && thisOption.totalCharge < thisExistingOption.totalCharge)
+    				if( ((sortType eq 'price' && thisOption.totalCharge < thisExistingOption.totalCharge)
     				  	||
-    					(sortType eq 'sortOrder' && thisOption.shippingMethodSortOrder < thisExistingOption.shippingMethodSortOrder) ) {
-
+    					(sortType eq 'sortOrder' && thisOption.shippingMethodSortOrder < thisExistingOption.shippingMethodSortOrder)) && !this.hasOption(optionsArray, thisOption)) {
+
```
rmarchandTen24: give hasOption priority in the if statement by putting it at the top

PR #5236, opened 2017-02-01
Replace IIF in slatwall with custom (non-string evaluating) function

gerickson808:
This change is [Reviewable](https://reviewable.io/reviews/ten24/slatwall/5236)


org/Hibachi/HibachiUtilityService.cfc
```diff
@@ -218,6 +218,10 @@
 			return result.toString().toUppercase();
   		}
 
+  		public any function hibachiTernary(required any condition, required any expression1, required any expression2){
+  			return (condition)? expression1 : expression2;
```
rmarchandTen24: @gerickson808 can you explicity reference the arguments scope
arguments.condition
arguments.expression1 etc.

PR #5243, opened 2017-02-03
Updates a method to give extra functionality.

ian-a-hickey:
This updates makes it so you can get images via width and height explicitly instead of just using the predefined sizes.

---
This change is [Reviewable](https://reviewable.io/reviews/ten24/slatwall/5243)


gregmoser:
@rmarchandTen24 and @ian-a-hickey this should not have been merged in.  You were already able to pass width & height.  Here is a long google group thread about it, and @rmarchandTen24 and I spent time looking at how it works.

https://groups.google.com/forum/#!topic/slatwallecommerce/7YAER8wHxgc

This PR actually REMOVED functionality.

PR #5261, opened 2017-02-14
removing reference to session

rmarchandTen24:
This change is [Reviewable](https://reviewable.io/reviews/ten24/slatwall/5261)

### DIFF
--- a/model/entity/Collection.cfc
+++ b/model/entity/Collection.cfc
@@ -105,8 +105,8 @@ component displayname="Collection" entityname="SlatwallCollection" table="SwColl
 	public any function init(){
 		super.init();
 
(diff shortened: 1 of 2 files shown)


sumitverma:
@rmarchandTen24 This wont work, because application is still server specific.

gregmoser:
Also this would cause MAJOR issues because everyone's collections would be in the same application scope... application.collection would get updated in mid-request if someone else instantiated a collection entity.

gregmoser:
shouldn't this use: HibachiCacheService.getCachedValue() / setCachedValue() / hasCachedValue() so that you can point multiple application servers to the same cache storage

sumitverma:
Goal is to have it to work out of the box with multiple application server
without having a shared cache storage.

Sumit Verma

PR #5273, opened 2017-02-15
Updated session to run through cacheService for centralization

gregmoser:


org/Hibachi/HibachiScope.cfc
```diff
@@ -31,102 +31,87 @@ component output="false" accessors="true" extends="HibachiTransient" {
...
-			sessionKey = COOKIE.CFID;
-		}
-		lock name="#sessionKey#_#getHibachiInstanceApplicationScopeKey()#_#arguments.key#" timeout="10" {
-			if(!structKeyExists(session, getHibachiInstanceApplicationScopeKey())) {
-				session[ getHibachiInstanceApplicationScopeKey() ] = {};
-			}
-			session[ getHibachiInstanceApplicationScopeKey() ][ arguments.key ] = arguments.value;
-		}
+		var cacheKey = getSession().getSessionCookieNPSID() & "-" & arguments.key;
+		getService("HibachiCacheService").setCachedValue(cacheKey, arguments.value);
```
sumitverma: This is not good because cacheservice is a singleton and this way you are now storing session related data in application cache. Concept of delegating back to cache service is good, but the key set with setSessionValue still needs to be session (shared or local) scope. And in Slatwall core we want to just get rid of any usage of setSessionValue to support LB out of the box.

gregmoser: If you notice the key itself is session specific and it comes from a persistent value that is stored in the database.  yes, all keys are stored in 1 central cache (either internal or external) but they are compartmentalized by a session key.

This PR isn't meant to be merged, it's meant for discussion, tag it however you like.

@rmarchandTen24 was going to set up a call so that the 3 of us so that I could discuss in more detail.  If you want to go ahead and put something on the books that would be great, there is a lot to talk about.  I also came up with a solution for multi-server locking for no duplicate orders, etc.

PR #5298, opened 2017-02-27
Account Payment Methods can be deleted even if they have order payments

jasperboyd:
relationship is severed before delete to prevent cascade

---
This change is [Reviewable](https://reviewable.io/reviews/ten24/slatwall/5298)


rmarchandTen24:
@jasperboyd We should be able delete account payment methods, so it's OK in core. But there should be validation to not allow deleting an account payment method assigned to subscription usage

model/validation/AccountPaymentMethod.json
```diff
@@ -19,7 +19,7 @@
 		"creditCardNumber":		[{"contexts":"save","conditions":"paymentTypeCreditCardAndNew,editExistingCreditCardNumber","required":true,"dataType":"creditCard"}],
 		"expirationMonth":		[{"contexts":"save","conditions":"paymentTypeCreditCard","required":true,"dataType":"numeric"}],
 		"expirationYear":		[{"contexts":"save","conditions":"paymentTypeCreditCard","required":true,"dataType":"numeric"}],
-		"orderPayments":		[{"contexts":"delete","maxCollection":0}],
+		"orderPayments":		[{"contexts":"delete","method":"hasNoSubscriptionUsage"}],
```
sumitverma: I think to keep this consistent you should change it to
"subscriptionUsages":		[{"contexts":"delete","maxCollection":0}],

PR #5323, opened 2017-03-14
Updates order.cfc to use val instead of len.

ian-a-hickey:
This is an issue I found while doing a custom project update.

---
This change is [Reviewable](https://reviewable.io/reviews/ten24/slatwall/5323)


gregmoser:
This changes the behavior and makes it so that if this value isn't set, then it says the order cannot be placed without receiving payment... that is incorrect.  If this setting is left blank then it is fine to take the order, only when there is a value do we need to check it.

PR #5335, opened 2017-03-17
API Settings for Dirty Read (Read Uncommitted) and Max Page Show

jasperboyd:
This change is [Reviewable](https://reviewable.io/reviews/ten24/slatwall/5335)

### DIFF
--- a/model/entity/Collection.cfc
+++ b/model/entity/Collection.cfc
@@ -604,6 +604,9 @@ component displayname="Collection" entityname="SlatwallCollection" table="SwColl
 		//Simple Filters
 		for (var key in data){
 			//handle filters.
+			if(!isSimpleValue(data[key])){
+ 				continue; 
+ 			} 
 			if (left(key, 3) == "fr:"){
 				
 				var prop = key.split(':')[2];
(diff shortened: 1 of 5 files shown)


model/entity/Collection.cfc
```diff
@@ -604,6 +604,9 @@ component displayname="Collection" entityname="SlatwallCollection" table="SwColl
 		//Simple Filters
 		for (var key in data){
 			//handle filters.
+			if(!isSimpleValue(data[key])){
```
rmarchandTen24: @jasperboyd @sumitverma  IMHO, we probably shouldn't be doing passover logic. It's similar to try catching. If someone puts in a bad data type, then they should be informed of the error.

jasperboyd: @rmarchandTen24 I agree that an error should be thrown in that case. However what's happening here is that data is actually request context and Hibachi's request setup logic is nesting params IE category.categoryID is building out a struct called category that contains categoryID even though the actually param is something like: ?f:category.categoryID:EQ=

We can definitely review this and see if there's a more robust solution for handling the problem, I thought that this would be acceptable because the logic is expecting to be working with strings so this will passover everything in request struct that isn't a string.

rmarchandTen24:
@jasperboyd @sumitverma we have one of these and could just make it general https://github.com/ten24/slatwall/pull/5335/files#diff-bbef4292a01371cc041840b5f398fec6R223

sumitverma:
I think having separate setting to API and smartlist makes sense

PR #5360, opened 2017-03-30
Develop team pick pack ship ih client

ian-a-hickey:
This is the first round of pull request for the client side orderFulfillment listing work.

---
This change is [Reviewable](https://reviewable.io/reviews/ten24/slatwall/5360)


admin/client/src/orderfulfillment/components/sworderfulfillmentlist.ts
```diff
@@ -0,0 +1,135 @@
...
+        }
+
+        var addFilter = (key, value) => {
+            console.log("Add Filter Called With", key, value);
+            if (key == "partial"){
+                this.orderFulfillmentCollection.addFilterGroup([{
+                    propertyIdentifier: "orderFulfillmentInventoryStatusType.typeID",
+                    comparisonValue: "fefc92c1d8184017aa65cdc882bdf637",
+                    comparisonOperator: "=",
+                    logicalOperator: "OR",
```
rmarchandTen24: so far looks good. You may or may not need to make the logical Operator "OR" part conditional if it is the only filter. I'm guessing you already have some filters.

admin/client/src/orderfulfillment/components/sworderfulfillmentlist.ts
```diff
@@ -0,0 +1,135 @@
...
+                this.orderFulfillmentCollection.addFilter("orderFulfillmentItems.stock.location.locationName", value);
+            }
+            this.refreshCollection();
+        }
+
+        /**
+         * Initialized the orderFulfillment collection so that the listingDisplay can you it to display its data.
+         */
+        this.refreshCollection = () => {
+            this.orderFulfillmentCollection.getEntity().then((response)=>{
```
rmarchandTen24: it's a good idea to put all promise requests as deferred promises inside of a service. This makes it easier for multiple areas to tap into the .then of a promise without making duplicate requests

admin/client/src/orderfulfillment/orderfulfillment.module.ts
```diff
@@ -0,0 +1,24 @@
...
+//models 
+
+var orderfulfillmentmodule = angular.module('orderFulfillment',[coremodule.name])
+.config([()=>{
+
+}]).run([()=>{
+
+}])
+//constants
+.constant('orderFulfillmentPartialsPath','orderFulfillment/components/')
```
rmarchandTen24: we should make all folders and javascript files lowercase to support AWS and linux

admin/client/src/orderfulfillment/components/sworderfulfillmentlist.ts
```diff
@@ -0,0 +1,388 @@
...
+             processObject['data']['orderItemIDList'] = this.listAppend(processObject['data']['orderItemIDList'], callBackData.selection)
+        }else{
+             processObject['data']['orderItemIDList'] = this.listRemove(processObject['data']['orderItemIDList'], callBackData.selection);
+        }
+    };
+
+    /**
+     * Add Instance Of string to list
+     */
+    public listAppend = (str, subStr) => {
```
rmarchandTen24: we should already have a list append function in the utilityService. Can you use that one?

admin/client/src/orderfulfillment/components/sworderfulfillmentlist.ts
```diff
@@ -0,0 +1,388 @@
...
+        if (subStr){
+            str = str + ((isNew)? "" : ",") + subStr;
+        }
+        return str;
+    }
+    
+    /**
+     * Removes a string from a string.
+     */
+     public listRemove = (str, subStr) => {
```
rmarchandTen24: would be cool to have this in utilityservice.ts as well

admin/client/src/orderfulfillment/components/sworderfulfillmentlist.ts
```diff
@@ -0,0 +1,388 @@
...
+            return;
+        }
+        
+        return this.collections[view];
+     }
+
+    /**
+     * Setup the initial orderFulfillment Collection.
+     */
+     private createOrderFulfillmentCollection = (collectionConfigService) => {
```
rmarchandTen24: if the collectionConfigService is injected as public then you should be able to access it via this.collectionConfigService.

admin/client/src/orderfulfillment/components/sworderfulfillmentlist.ts
```diff
@@ -0,0 +1,388 @@
...
+        this.orderFulfillmentCollection.addDisplayProperty("shippingAddress.stateCode");
+        this.orderFulfillmentCollection.addDisplayProperty("orderFulfillmentInvStatusType.typeName");
+        this.orderFulfillmentCollection.addFilter("orderFulfillmentStatusType.typeName", "Fulfilled", "!=");
+        this.orderFulfillmentCollection.addFilter("order.orderNumber", null, "!=");
+     }
+    
+    /**
+     * Setup the initial orderItem Collection.
+     */
+    private createOrderItemCollection = (collectionConfigService) => {
```
rmarchandTen24: same here. Just use the property on the class

admin/client/src/orderfulfillment/components/sworderfulfillmentlist.ts
```diff
@@ -0,0 +1,388 @@
...
+                    this.total = response.recordsCount;
+                }
+            });
+        }
+    }
+
+    /**
+     * Redirects the current page (to go to login) if the user tries to interacts with the view while not logged in.
+     */
+    public redirect = () => {
```
rmarchandTen24: cool stuff. We might be able to handle this via the hibachi interceptor which applies to all requests for the multipage sections of the app

org/Hibachi/HibachiControllerREST.cfc
```diff
@@ -156,6 +156,18 @@ component output="false" accessors="true" extends="HibachiController" {
         }
 
     }
+    
+    public void function doProcess (required struct rc) {
```
rmarchandTen24: consolidate this to use the post method which i think is accessible in hibachiservice.ts

org/Hibachi/client/src/core/components/swtypeaheadinputfield.ts
```diff
@@ -76,6 +76,7 @@ class SWTypeaheadInputField implements ng.IDirective{
 	public scope = {};
 
 	public bindToController = {
+        name:"@",
```
rmarchandTen24: remove this becaue we can use fieldname

admin/client/src/orderfulfillment/components/orderfulfillmentlist.html
```diff
@@ -0,0 +1,502 @@
...
+				<div class="col-sm-6 col-md-6 col-lg-2 s-detail-module">
+			        
+					<!--Small content block-->
+			        <div class="s-sm-content-block">
+						
+						<div class="dropdown s-action">
+							<a href="##" class="dropdown-toggle" id="changeThisLabel1" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false">
+								<i class="fa fa-ellipsis-h" title="actions"></i>
+							</a>
+							<ul class="dropdown-menu pull-right" aria-labelledby="switchesView" ng-if="swOrderFulfillmentListController.getView() == swOrderFulfillmentListController.views.Fulfillments" ng-click="swOrderFulfillmentListController.setView(swOrderFulfillmentListController.views.Items)">
```
rmarchandTen24: here is an example of where a function should not be used in a template. Functions should be reserved for events and not binds because they cannot be properly dirty checked and therefore run every digest cycle.
ng-click is cool to use a function but keep them out of the bindings and structural directives. @ian-a-hickey can you refactor the functions out of the binds accross the templates

ian-a-hickey:
@rmarchandTen24 I fixed the merge conflicts, etc. Waiting on checks then this should be good.

PR #5369, opened 2017-04-05
added calculated fields to order item

sumitverma:
This change is [Reviewable](https://reviewable.io/reviews/ten24/slatwall/5369)


sumitverma:
@rmarchandTen24 length issue should be fixed now.

PR #5383, opened 2017-04-17
Develop team pick pack ship ih client code review updates

ian-a-hickey:
Updates the listingDisplays to use a direct reference to a property instead of a function that returns the same.
Removed the passing in of the collectionConfigService instead of referencing the polymorphic this.
Removed now unused getters.
Added a type for the collectionFilterValues I'm using.
Updates the name of the orderFulfillmentService to remove the SW from the name.
Updates the module path for the service to make lowercase.
Updates the package.json to reference jasmine-core
Updated the swtypeahead to remove the added binding for name (using fieldname instead).
Updated the templates to change the getView() to view in all instances.

---
This change is [Reviewable](https://reviewable.io/reviews/ten24/slatwall/5383)


rmarchandTen24:
@ian-a-hickey can you resolve the conflicts in this? You should probably stop committing the bundle to slatwall prs

rmarchandTen24:
@ian-a-hickey can you investigate why unit tests are failing?

PR #5404, opened 2017-04-24
Develop team pick pack ship client location path

ian-a-hickey:
This updates the location to be a calculated property. This also adds a migration script for the locationIdPath to be built. Tested working locally. This also updates the typeaheads on the orderFulfillmentListing to display the calculated property.

---
This change is [Reviewable](https://reviewable.io/reviews/ten24/slatwall/5404)


ian-a-hickey:
Updated this to use locationPathName for the simple rep instead of looping.

PR #5405, opened 2017-04-24
Adds the updates for the orderitems listing into core.

ian-a-hickey:
If there is an address on the orderReturn because this is a return item, show that item in the orderItems listing instead of trying to show the orderFulfillment shipping (which won't exist). This relates to this task:  https://ten24.teamwork.com/#/tasks/12195438?c=5901084 where it was found that this address is not being displayed.

---
This change is [Reviewable](https://reviewable.io/reviews/ten24/slatwall/5405)


admin/client/src/orderitem/components/sworderitems.ts
```diff
@@ -87,29 +87,43 @@ class SWOrderItems{
...
+ 						 ,sku.subscriptionTerm.subscriptionTermName
+ 						 ,sku.imageFile
+                         ,stock.location.locationName`
  					   
                       )
  					.addFilter('order.orderID',scope.orderId)
  					.addFilter('parentOrderItem','null','IS')
  					.setKeywords(scope.keywords)
  					.setPageShow(scope.paginator.getPageShow())
-					.setCurrentPage(scope.paginator.getCurrentPage())
```
rmarchandTen24: @ian-a-hickey  how come this is being removed?

PR #5413, opened 2017-04-25
getting boolean values to start importing as 0, 1 and fixing null case for empty strings

jasperboyd:
This change is [Reviewable](https://reviewable.io/reviews/ten24/slatwall/5413)

### DIFF
--- a/org/Hibachi/HibachiDAO.cfc
+++ b/org/Hibachi/HibachiDAO.cfc
@@ -327,9 +327,11 @@
 							#arguments.tableName#
 						SET
 							<cfloop from="1" to="#listLen(keyList)#" index="local.i">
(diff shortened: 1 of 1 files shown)


org/Hibachi/HibachiDAO.cfc
```diff
@@ -327,7 +327,11 @@
 							#arguments.tableName#
 						SET
 							<cfloop from="1" to="#listLen(keyList)#" index="local.i">
-								<cfif arguments.updateData[ listGetAt(keyList, i) ].value eq "NULL" OR (arguments.insertData[ listGetAt(keyList, i) ].value EQ "" AND arguments.insertData[ listGetAt(keyList, i) ].dataType EQ "timestamp")>
```
rmarchandTen24: does this still handle the timestamp case? Also can you rectangularize the code?

jasperboyd: The timestamp case was to insert null for empty string timestamps. But in the case of importing a CSV where all null values are set to empty string we want to substitute null. 

And I've updated it with rectangularized version

PR #5414, opened 2017-04-25
giving validate csv function flexibility to override whether the passed path is expanded or not

jasperboyd:
This change is [Reviewable](https://reviewable.io/reviews/ten24/slatwall/5414)

### DIFF
--- a/org/Hibachi/HibachiDataService.cfc
+++ b/org/Hibachi/HibachiDataService.cfc
@@ -63,15 +63,18 @@ component accessors="true" output="false" extends="HibachiService" {
 		return csvQuery; 
 	}	
 
-	public array function validateCSVFile(required string pathToCSV, string expectedColumnHeaders){
-		var csvFile = FileOpen(pathToCSV);
+	public array function validateCSVFile(required string pathToCSV, string expectedColumnHeaders, boolean expandedPath = false){
+		if(!arguments.expandedPath){ 
+			arguments.pathToCSV = ExpandPath(arguments.pathToCSV); 
+		} 
+		var csvFile = FileOpen(arguments.pathToCSV);
 		var i = 1;
 		var problemLines = []; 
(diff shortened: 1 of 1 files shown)


org/Hibachi/HibachiDataService.cfc
```diff
@@ -63,15 +63,18 @@ component accessors="true" output="false" extends="HibachiService" {
...
+		} 
+		var csvFile = FileOpen(arguments.pathToCSV);
 		var i = 1;
 		var problemLines = []; 
 		while(!FileisEOF(csvFile)){ 
 			var line = FileReadLine(csvFile);
 			if(i == 1){
 				var columnsList  = REReplaceNoCase(line, "[^a-zA-Z\d,]", "", "all");
-				if(structKeyExists(arguments, "expectedColumnHeaders")){
+				if(structKeyExists(arguments, "expectedColumnHeaders") && len(arguments.expectedColumnHeaders) > 0){
```
rmarchandTen24: no need to make a change but you can simplify this to just len(arguments.expectedColumnHeaders). No need to check if it's greater than 0. because it will evaluate as true or false by the number itself already

org/Hibachi/HibachiDataService.cfc
```diff
@@ -63,15 +63,18 @@ component accessors="true" output="false" extends="HibachiService" {
 		return csvQuery; 
 	}	
 
-	public array function validateCSVFile(required string pathToCSV, string expectedColumnHeaders){
-		var csvFile = FileOpen(pathToCSV);
+	public array function validateCSVFile(required string pathToCSV, string expectedColumnHeaders, boolean expandedPath = false){
```
rmarchandTen24: This works but in my opinion the function is gaining some unnecessary configuration and starting to do more work by doing this. IMO it would be better to just use expand path before passing in the path to the function

PR #5422, opened 2017-04-26
Top nav don't need angular bootstrap to start working

migueltarga:
@rmarchandTen24 This will fix the problem with our navbar dropdowns.
We have bootstrap.js and jquery in Slatwall, we don't need to wait for the Angular bootstrap to get it to work.

---
This change is [Reviewable](https://reviewable.io/reviews/ten24/slatwall/5422)


alamcordeiro:
Excelent

PR #5434, opened 2017-05-02
Move some logic further into a method.

ian-a-hickey:
An issue was found in a custom project where the shipping from an international rate was returning a default value before any address was associated with the cart. The issue was the line below, all returned true in the case where the int shipping method was using split shipment. Moving that condition to the end of the function allows the other constrains to work before getting there. In this case, it will fail the address check and not return a value until there is an address.

---
This change is [Reviewable](https://reviewable.io/reviews/ten24/slatwall/5434)

### DIFF
--- a/model/service/ShippingService.cfc
+++ b/model/service/ShippingService.cfc
@@ -577,10 +577,6 @@ component extends="HibachiService" persistent="false" accessors="true" output="f
 		// Make sure that the rate is active
 		if(!isNull(shippingMethodRate.getActiveFlag()) && isBoolean(shippingMethodRate.getActiveFlag()) && !shippingMethodRate.getActiveFlag()) {
 			return false;
-		}
-
-		if(!isNull(shippingMethodRate.getSplitShipmentWeight())){
-			return true; 
 		} 
 
 		// Make sure that the orderFulfillment Item Price is within the min and max of rate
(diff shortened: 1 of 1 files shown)


model/service/ShippingService.cfc
```diff
@@ -651,6 +647,12 @@ component extends="HibachiService" persistent="false" accessors="true" output="f
 			return false;
 		}
 		
+		//This check must happen after the address check.
+ 		if(!isNull(shippingMethodRate.getSplitShipmentWeight())){
```
rmarchandTen24: @ian-a-hickey I'm not sure this adds any value at this point. because if it evaluates as false then it passes through and hits true anyway. Did you actually intend this to return false if splitShipmentWeight is null under a certain condition?

ian-a-hickey: I'm not sure I understand what you mean. That logic is not new, it was just moved frmo the beginning to the end of the containing function. It has to happen after the other constraint checks. This logic must happen after the other constraints, otherwise there is no point in having constraints in the first place, as they will never be evaluated if this is a split shipment enabled method.  Understand? This logic as is, is saying, if this method has split shipment set, its always ok to return it. Which of course is not true. It should only return if it has the correct address or quantity etc. On the custom project it was returning a huge default amount as soon as the item what added to the cart. If you look here: https://github.com/ten24/slatwall/blob/2842c043aa619cdd3469228335de7bd9bb53685f/model/service/ShippingService.cfc#L361 which is what is calling this method it says " If we should have gotten a response bean from the shipping integration but didn't then use the default amount if and only if we should be using this in the first place." The problem I fixed is that it always returns true when using split shipment. Which fixes the issue because now it does return the value until it meets the requirements for returning it. @jasperboyd may be able to explain better.

ian-a-hickey: @rmarchandTen24 On closer inspection, I understand. The logic doesn't need to be there in the first place. Its not that moving it is bad, just that it doesn't do anything. I agree, I'll remove it.

ian-a-hickey: @rmarchandTen24 @jasperboyd This is going to require more work. Basically, Jasper was returning true to skip portions of the logic related to weight constraints. The easy fix is only doing the weight check when its not a split ship, but will require testing. Putting this on hold for now.

PR #5435, opened 2017-05-02
Added max file size attribute option

gerickson808:
This change is [Reviewable](https://reviewable.io/reviews/ten24/slatwall/5435)


model/entity/HibachiEntity.cfc
```diff
@@ -174,7 +174,7 @@ component output="false" accessors="true" persistent="false" extends="Slatwall.o
 		return variables.attributeValuesForEntity;
 	}
 
-	public any function getAttributeValue(required string attribute, returnEntity=false){
+	public any function getAttributeValue(required string attribute, returnEntity=false, usePropertyIfExists=true){
```
rmarchandTen24: @gerickson808 we probably don't need to add another argument to get this working and it would be better not to have properties that sound like conditionals. We should go over this tomorrow

PR #5399, opened 2017-04-20
Implement "Copy Promotion Period" Functionality

chriskent:
Implement "Copy Promotion Period" Functionality

---
This change is [Reviewable](https://reviewable.io/reviews/ten24/slatwall/5399)

PR #5442, opened 2017-05-04
Add clone account process

chriskent:
This change is [Reviewable](https://reviewable.io/reviews/ten24/slatwall/5442)


model/service/AccountService.cfc
```diff
@@ -382,6 +382,112 @@ component extends="HibachiService" accessors="true" output="false" {
...
+				getHibachiDAO().save(accountAuthentication);
+	
+				// Set the password
+				accountAuthentication.setPassword( getHashedAndSaltedPassword(arguments.processObject.getPassword(), accountAuthentication.getAccountAuthenticationID()) );
+			}
+
+			// Clone account addresses
+			if(!isNull(processObject.getCloneAccountAddressesFlag())) {
+				for(var accountAddress in arguments.account.getAccountAddresses()){
+					var newAccountAddress = duplicate(accountAddress);
```
sumitverma: @rmarchandTen24 @chriskent We should not use duplicate on objects. Duplicate does a deep copy, that could lead to recursion.

chriskent: Hi Sumit,
What is the preferred and safe method to duplicate an object in Slatwall?

sumitverma: Manually setting each property. That will also make sure you are not duplicating unwanted properties, i.e. created/modified by etc.

At some point we had talked about creating a Hibachi Duplicate function which you can pass the property list, but I'm not sure if that was ever done.

chriskent: OK. I will make this change to manually setting each require property for this PR and also https://github.com/ten24/slatwall/pull/5399 where I had also used duplicate.

It would be good to have a Hibachi function that could safely duplicate an object including all properties with the except certain properties e.g. created/modified etc. As this would handle cases where the new properties are added to the entity in the future.

PR #5449, opened 2017-05-09
Api:main.post delete response fix

gerickson808:
This change is [Reviewable](https://reviewable.io/reviews/ten24/slatwall/5449)


org/Hibachi/HibachiControllerREST.cfc
```diff
@@ -749,7 +752,7 @@ component output="false" accessors="true" extends="HibachiController" {
 	            }
 	        }
 
-	        if(entity.hasErrors()){
+	        if(entity.hasErrors() || (!isNull(deleteOK) && deleteOK == false)){
```
rmarchandTen24: @gerickson808 this part of code is now expecting deleteOK however it is only delcared in the delete context. If the delete process is run then the entity will have errors anyway so you probably shouldn't even have to do this. You should be able to remove " || (!isNull(deleteOK) && deleteOK == false)".

PR #5456, opened 2017-05-12
This adds orderitem validation condition on quantity to only fire if the

ian-a-hickey:
order is not placed or if the quantity is being changed. This fixes an issue in a custom project.

---
This change is [Reviewable](https://reviewable.io/reviews/ten24/slatwall/5456)


model/entity/OrderItem.cfc
```diff
@@ -469,6 +483,9 @@ component entityname="SlatwallOrderItem" table="SwOrderItem" persistent="true" a
 	}
 
 	public void function setQuantity(required numeric quantity){
+		if (isNull(variables.quantity) || arguments.quantity != variables.quantity){
```
rmarchandTen24: you should just add a default=0 on the property so we won't have to check structkeyexists or isNull on variables.quantity

rmarchandTen24: I think that the isNull check is only going to work if getQuantity was called before setQuantity.
structkeyexists would be used before going to the database
isNull for after

ian-a-hickey: This is updated.

model/entity/OrderItem.cfc
```diff
@@ -48,7 +48,7 @@ component entityname="SlatwallOrderItem" table="SwOrderItem" persistent="true" a
 	property name="price" ormtype="big_decimal";
 	property name="skuPrice" ormtype="big_decimal";
 	property name="currencyCode" ormtype="string" length="3";
-	property name="quantity" hb_populateEnabled="public" ormtype="integer";
+	property name="quantity" hb_populateEnabled="public" ormtype="integer" default="0";
```
rmarchandTen24: Sorry for the confusion. Add the default to the non persistent property quantity has changed and not quantity itself

PR #5473, opened 2017-05-19
setting running flag to false for workflows that have been running for longer than 90 minutes

jasperboyd:
This change is [Reviewable](https://reviewable.io/reviews/ten24/slatwall/5473)


rmarchandTen24:
@jasperboyd we should probably set runningFlag to false when the try catch fails. 
https://github.com/ten24/slatwall/pull/5473/files#diff-f885d5e65e85cc097533678a4e95d4e4R231

jasperboyd:
@rmarchandTen24 We already do that https://github.com/ten24/slatwall/blob/master/model/service/WorkflowService.cfc#L226

PR #5496, opened 2017-05-31
changed line 82 where 'currency' was being entered as an argument to …

Pete-Eichman:
…getAmount(). changed the argument from 'currency' to 'percentage'.

---
This change is [Reviewable](https://reviewable.io/reviews/ten24/slatwall/5496)

### DIFF
--- a/admin/views/entity/promotionrewardtabs/currencies.cfm
+++ b/admin/views/entity/promotionrewardtabs/currencies.cfm
@@ -2,45 +2,45 @@
 
     Slatwall - An Open Source eCommerce Platform
     Copyright (C) ten24, LLC
-	
+
     This program is free software: you can redistribute it and/or modify
     it under the terms of the GNU General Public License as published by
     the Free Software Foundation, either version 3 of the License, or
     (at your option) any later version.
-	
+
     This program is distributed in the hope that it will be useful,
     but WITHOUT ANY WARRANTY; without even the implied warranty of
     MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
     GNU General Public License for more details.
-	
+
     You should have received a copy of the GNU General Public License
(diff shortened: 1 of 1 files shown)


admin/views/entity/promotionrewardtabs/currencies.cfm
```diff
@@ -75,18 +75,18 @@ Notes:
 				<cfif structKeyExists( local.promoRewardCurrencyStruct, local.currency.getCurrencyCode())>
 					<td>
 						#$.slatwall.formatValue(local.promoRewardCurrencyStruct[ local.currency.getCurrencyCode() ].getAmount(), 'currency', {currencyCode=local.currency.getCurrencyCode()} )#
-						
+
 					</td>
 				<cfelseif local.currency.getCurrencyCode() eq rc.promotionReward.getCurrencyCode()>
 					<td>
-						#$.slatwall.formatValue(rc.promotionReward.getAmount(), 'currency', {currencyCode=local.currency.getCurrencyCode()} )#
+						#$.slatwall.formatValue(rc.promotionReward.getAmount(), 'percentage', {currencyCode=local.currency.getCurrencyCode()} )#
```
sumitverma: If we are passing currenycode, why are we formatting as percentage? Also, please make all pull requests to hotfix or develop.

rmarchandTen24: @Pete-Eichman @sumitverma  i think the issue is based around the header here https://github.com/ten24/slatwall/pull/5496/files#diff-f019114c8247da2b2c9a09e31809bd5dR63.

@Pete-Eichman if you look at this function https://github.com/ten24/slatwall/blob/master/model/entity/PromotionReward.cfc#L125

you can see the different amount types which could be applied to the persisted Promotion Reward.

those being percentageOff, amount, and amountOff

what you really want to do is make amount and amountOff formatted with currency and percentageOff formatted with percentage

PR #5498, opened 2017-05-31
Removed attribute loop from order item collection on order detail

gerickson808:
Apparently attributes aren't supported by collections, so when there were any attributes for order items this would break order item retrieval.

---
This change is [Reviewable](https://reviewable.io/reviews/ten24/slatwall/5498)

### DIFF
--- a/admin/client/src/orderitem/components/sworderitems.ts
+++ b/admin/client/src/orderitem/components/sworderitems.ts
@@ -127,17 +127,6 @@ class SWOrderItems{
 					.setCurrentPage(scope.paginator.getCurrentPage())
  					;
 
-					//add attributes to the column config
-					angular.forEach(scope.attributes,function(attribute){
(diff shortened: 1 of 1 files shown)


admin/client/src/orderitem/components/sworderitems.ts
```diff
@@ -127,17 +127,6 @@ class SWOrderItems{
 					.setCurrentPage(scope.paginator.getCurrentPage())
  					;
 
-					//add attributes to the column config
```
rmarchandTen24: @gerickson808 I'm pretty sure we still want to show all related attributes for the orderItem on the listing. @sumitverma is this something we want to change? Are we running into performance issues as a result?

gerickson808: The problem is that it's a collection, which no longer support attributes.

PR #5511, opened 2017-06-07
Develop ih workflow

ian-a-hickey:
A bunch of small updates to workflow to fix little bugs. This pull request includes fixes for:
The task detail was not displaying information. 
The task detail was not updating on save (so if you changed the name, you couldn't see that until page refresh).
The trigger history tab was not filtering by the current workflow. Instead, it was showing the trigger history for all workflows.
This also has a fix for imageDetailModalLauncher because it was causing an error in the bundle. I made two values strings instead of integers.
This fixes productCreate never being announced and not having a way to key into it in workflow. I added create to the event options, and then added an announcement for the process during the save by checking if it is new. I tested it by creating a product and then checking the trigger history.
This also fixes an issue where the task condition was not being saved on edit.

---
This change is [Reviewable](https://reviewable.io/reviews/ten24/slatwall/5511)


org/Hibachi/HibachiService.cfc
```diff
@@ -183,19 +183,33 @@
...
 	        if(!arguments.entity.hasErrors()) {
 	            arguments.entity = getHibachiDAO().save(target=arguments.entity);
         
                 // Announce After Events for Success
 				getHibachiEventService().announceEvent("after#arguments.entity.getClassName()#Save", arguments);
 				getHibachiEventService().announceEvent("after#arguments.entity.getClassName()#SaveSuccess", arguments);
+				
+				//If new need to announce the Create process as well as Success
+				if (isNew){
+					getHibachiEventService().announceEvent("after#arguments.entity.getClassName()#Process_Create", arguments);
```
sumitverma: Event should be afterXXXCreate not afterXXXProcessCreate. We dont need "Process" in these events.

ian-a-hickey: I fixed this and retested.

PR #5526, opened 2017-06-13
Update the nodatasource error to return a 500

gregmoser:
This is instead of the 200 it returns now.

---
This change is [Reviewable](https://reviewable.io/reviews/ten24/slatwall/5526)


rmarchandTen24:
https://github.com/ten24/slatwall/pull/5567

PR #5566, opened 2017-06-28
ability to use subdomains with slatwall cms sites 

jasperboyd:
This change is [Reviewable](https://reviewable.io/reviews/ten24/slatwall/5566)


model/transient/HibachiScope.cfc
```diff
@@ -80,6 +80,13 @@ component output="false" accessors="true" extends="Slatwall.org.Hibachi.HibachiS
 	
 	public any function getCurrentRequestSite() {
 		var domain = listFirst(cgi.HTTP_HOST,':');
+		var subdomain = listFirst(cgi.HTTP_HOST,'.');
```
rmarchandTen24: @sumitverma @jasperboyd  I can see this becoming an issue when two sites share the same subdomain. Should the domain become its own object or just be part of the domainslist?

jasperboyd: @rmarchandTen24 that's a good point we should probably match on the proper domain name as well

sumitverma:
I don't think multiple site using the same domain will work. I can't think of a use case. We are not supporting the use of site code in the URL.

model/transient/HibachiScope.cfc
```diff
@@ -80,6 +80,46 @@ component output="false" accessors="true" extends="Slatwall.org.Hibachi.HibachiS
 	
 	public any function getCurrentRequestSite() {
 		var domain = listFirst(cgi.HTTP_HOST,':');
+		var urlArray = listToArray(listFirst(listFirst(cgi.HTTP_HOST,'/'),':'),'.'); 
+		var www = ''; 
+		var subdomain = ''; 
+		var domainName = ''; 
+		var topLevelDomain = urlArray[arrayLen(urlArray)]; 
+		
+		if(lcase(urlArray[1]) == "www"){
```
rmarchandTen24: @jasperboyd maybe looping in reverse would simplify this.
like you know the last item is always the extension .com
then domain
then subdomain or www

jasperboyd: @rmarchandTen24 in productions thats true but this also needs to work with are local configurations thats why I added the check to see if `topLevelDomain != domain` below

rmarchandTen24: ok let's get some tests to prove the scenarios going forward

jasperboyd: @rmarchandTen24 test in place and build is passing

meta/tests/unit/transient/SlatwallScopeTest.cfc
```diff
@@ -246,6 +246,69 @@ component extends="Slatwall.meta.tests.unit.SlatwallUnitTestBase" {
 		
 		assertFalse( structKeyExists(cd, "hushbuppy") );
 	}
+		
+	/**
+	* @test
```
rmarchandTen24: @jasperboyd the test wasn't running because it was missing the @test meta data. After I added that, it is running. However, it shows that it fails.

PR #5596, opened 2017-07-06
PublicService.ts tweaks, swTypeaheadInputField search endpoint added

gerickson808:
Also added transclusion to swAddressForm.

This change is [Reviewable](https://reviewable.io/reviews/ten24/slatwall/5596)


org/Hibachi/client/src/core/components/swtypeaheadsearch.ts
```diff
@@ -165,7 +167,9 @@ class SWTypeaheadSearchController {
         this.typeaheadService.setTypeaheadState(this.typeaheadDataKey, this);
 
         this.observerService.attach(this.clearSearch, this.typeaheadDataKey + 'clearSearch');
-	}
+
+        this.$http = $http;
```
ian-a-hickey: when using public / private in the constructor argument, you don't need to set the scope to this, its implicit. If you removed the private from constructor(...private $http), then you would need to. Doesn't hurt anything though.

gerickson808:
Oh yeah I didn't touch that. I copy-pasted the file over from Lightspeed to make sure I got all my changes and it screwed up the whitespace I think.
          
            
            On Jul 7 2017, at 9:42 am, Ian Hickey <notifications@github.com> wrote:
            
            
              @ian-a-hickey commented on this pull request.



In org/Hibachi/client/src/core/components/swtypeaheadsearch.ts:
> @@ -165,7 +167,9 @@ class SWTypeaheadSearchController {
         this.typeaheadService.setTypeaheadState(this.typeaheadDataKey, this);
 
         this.observerService.attach(this.clearSearch, this.typeaheadDataKey + 'clearSearch');
-	}
+
+        this.$http = $http;

when using public / private in the constructor argument, you don't need to set the scope to this, its implicit. If you removed the private from constructor(...private $http), then you would need to. Doesn't hurt anything though.

PR #5599, opened 2017-07-14
Develop ih collection config validation

ian-a-hickey:
Adds a number of validations to collections dealing with the collection config. This is not a complete list, but it should be a good start. I've done basic testing to validate that collections are still working and I'm going to add unit tests into this branch.

---
This change is [Reviewable](https://reviewable.io/reviews/ten24/slatwall/5599)


sumitverma:
I think from structure perspective, it may make sense to just have one validation rule isValidCollectionConfig which calls the isValidCollectionConfig method in entity which delegates all the checks. Thoughts?

PR #5621, opened 2017-07-28
Hotfix ih product bundles

ian-a-hickey:
Fixes a bunch of small issues that were breaking product bundles.
The validation was not working because it was throwing a error before the user entered any data.
The typeahead a not displaying.
The min and max were not updating on save (added a refresh function that doesn't need to reach out again to server).
Tested adding a new bundle. Adding 6 bundle groups with various configurations.


admin/client/src/productbundle/components/productbundlegroups.html
```diff
@@ -4,6 +4,7 @@
                 data-product-bundle-groups="swProductBundleGroups.productBundleGroups"
 		    		data-product-bundle-group="productBundleGroup"
 		    		data-index="$index"
+		    	data-refresh-product-bundle-group="swProductBundleGroups.refreshProductBundleGroup()"
```
rmarchandTen24: if you are just passing a function into scope you shouldn't call it. Try passing swProductBundleGroups.refreshProductBundleGroup instead of swProductBundleGroups.refreshProductBundleGroup()

ian-a-hickey: Agree. Both the others going into that diective were like that though for addProductBundleGroup and removeProductBundleGroup so figured there was a reason for it:
data-add-product-bundle-group="swProductBundleGroups.addProductBundleGroup()"
data-remove-product-bundle-group="swProductBundleGroups.removeProductBundleGroup(index)"

PR #5675, opened 2017-08-24
Split checkout billing address component out from individual payment methods

gerickson808:
This change is [Reviewable](https://reviewable.io/reviews/ten24/slatwall/5675)


ian-a-hickey:
This looks good to me.

PR #5699, opened 2017-09-05
Develop ih collection cache

ian-a-hickey:
This is what I had added when testing object level caching. I refactored it to use hibachiCacheService instead of direct cacheGet cachePuts. By updating the stateStruct to use collectionConfigStruct, its able to generate a unique name per transient collection for the frontend. Those transient collections will only be cached if setCacheable(true) when setting up the collection in the template.

---
This change is [Reviewable](https://reviewable.io/reviews/ten24/slatwall/5699)


rmarchandTen24:
@ian-a-hickey  @KevinJason just wanted to check to make sure you are able to wrap up the changes we talked about in this pr.

PR #5714, opened 2017-09-12
Swinput Hotfix

ian-a-hickey:
This fixes the input not working on lucee because the value is not returned as a string so the trim fails.

---
This change is [Reviewable](https://reviewable.io/reviews/ten24/slatwall/5714)

### DIFF
--- a/org/Hibachi/client/src/form/components/swinput.ts
+++ b/org/Hibachi/client/src/form/components/swinput.ts
@@ -97,7 +97,7 @@ class SWInputController{
 
 		if(this.metadataService.isAttributePropertyByEntityAndPropertyIdentifier(this.object,this.propertyIdentifier)){
 			this.object.validations.properties[name] = [];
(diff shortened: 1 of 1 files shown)


org/Hibachi/client/src/form/components/swinput.ts
```diff
@@ -97,7 +97,7 @@ class SWInputController{
 
 		if(this.metadataService.isAttributePropertyByEntityAndPropertyIdentifier(this.object,this.propertyIdentifier)){
 			this.object.validations.properties[name] = [];
-			if(this.object.metaData[this.propertyIdentifier].requiredFlag && (this.object.metaData[this.propertyIdentifier].requiredFlag.trim().toLowerCase()=="yes" || this.object.metaData[this.propertyIdentifier].requiredFlag == "true")){
+			if((this.object.metaData[this.propertyIdentifier].requiredFlag && this.object.metaData[this.propertyIdentifier].requiredFlag == true) || typeof this.object.metaData[this.propertyIdentifier].requiredFlag === 'string' && this.object.metaData[this.propertyIdentifier].requiredFlag.trim().toLowerCase()=="yes"){
```
rmarchandTen24: @ian-a-hickey  this is using java. prob should use lcase

PR #5716, opened 2017-09-12
Develop team pick pack ship filters and tests

ian-a-hickey:
pick pack into develop.

---
This change is [Reviewable](https://reviewable.io/reviews/ten24/slatwall/5716)


ian-a-hickey:
I tested this yesterday. I was able to create a new fulfillment batch etc. There may be a couple random issues from the merge, but so far it looks good.

PR #5725, opened 2017-09-15
default image functionality for option

jasperboyd:


model/entity/Image.cfc
```diff
@@ -54,6 +54,8 @@ component displayname="Image" entityname="SlatwallImage" table="SwImage" persist
 	property name="imageDescription" ormtype="string" length="4000" hb_formFieldType="wysiwyg";
 	property name="imageFile" ormtype="string" hb_formFieldType="file" hb_fileUpload="true" hb_fileAcceptMIMEType="image/gif,image/jpeg,image/pjpeg,image/png,image/x-png" hb_fileAcceptExtension=".jpeg,.jpg,.png,.gif";
 	property name="directory" ormtype="string";
+
+	property name="calculatedImagePath" ormtype="string";
```
rmarchandTen24: @jasperboyd, @sumitverma  this one is tricky because it depends on the baseimageurl in hibachiscope, which then depends on a setting. If the setting is changed then we would need to update all the calculatedImagePaths too. 

public string function getBaseImageURL() {
		return getURLFromPath(setting('globalAssetsImageFolderPath'));
	}

jasperboyd: @rmarchandTen24  Yeah that makes sense, maybe we write a hook for when that setting changes to update all image paths? Could be accomplished with a single query

PR #5727, opened 2017-09-15
Product Option Add To Cart Selector

jasperboyd:


model/entity/Product.cfc
```diff
@@ -341,11 +341,11 @@ component displayname="Product" entityname="SlatwallProduct" table="SwProduct" p
 	public array function getOptionGroups() {
```
rmarchandTen24: this returns a different data type than the original function. The original function returns objects and now it returns data. This will certainly cause a regression for anything depending on this function and it is also breaks convention in describing what it returns.

jasperboyd: @rmarchandTen24 / @sumitverma  we could setup a sister function here that returns the collection version of the data I was under the impression we were beginning to migrate anywhere using a smart list to be a collection

sumitverma: We do want to migrate away from smartlist, but in a backward compatible way so it doesn't break existing code.

jasperboyd:
@rmarchandTen24 changes made there's still going to be at least one error in ci related to `calculatedImagePage` from the other PR

rmarchandTen24:
@jasperboyd must please the circle ci god. It demands rbkeys

jasperboyd:
@rmarchandTen24 see my note above this code depends on the calculated property for `calculatedImagePath` in this pr that's the parts that's failing https://github.com/ten24/slatwall/pull/5725

PR #5735, opened 2017-09-20
fixing issue with options not being returned for workflowObject in create workflow

jasperboyd:
This change is [Reviewable](https://reviewable.io/reviews/ten24/slatwall/5735)


rmarchandTen24:
@jasperboyd can you make this change in hotfix. I think it's broken in stable

jasperboyd:
Yep @rmarchandTen24 you're right it was there as well https://github.com/ten24/slatwall/pull/5737

PR #5738, opened 2017-09-20
Bug ambiguous entity

UltimoTG:
When there’s a custom cfc in the integrationServices.[packageName].model.entity folder, SW throws an “ambiguous bean” error at startup

---
This change is [Reviewable](https://reviewable.io/reviews/ten24/slatwall/5738)


rmarchandTen24:
@UltimoTG this copies the integration beans into model entity. This is why we don't have cfclocation pointing to integrationservices anymore.

PR #5739, opened 2017-09-21
Bug fixes for makeup breakup and checkout

gerickson808:
This change is [Reviewable](https://reviewable.io/reviews/ten24/slatwall/5739)


public/views/templates/partials/checkout/addressform.html
```diff
@@ -1,169 +0,0 @@
-<!-- Address Form Partial - You can specify a prefix for each address for the name property of each field on the form -->
```
rmarchandTen24: @gerickson808 did you intend to delete this?

gerickson808: Yeah, it shouldn't exist by default.

gerickson808: The one the directive uses is in org/hibachi

PR #5792, opened 2017-10-05
Added recordAlias support after pagination for hibachiListingDisplay

gerickson808:
@rmarchandTen24 This is related to make up break up functionality.

---
This change is [Reviewable](https://reviewable.io/reviews/ten24/slatwall/5792)


admin/controllers/ajax.cfc
```diff
@@ -218,6 +218,9 @@ component persistent="false" accessors="true" output="false" extends="Slatwall.o
 					if(structKeyExists(admin,"processActionProperty")){
 						var processActionProperty=listlast(admin.processActionProperty,'.');
 						var processActionPropertyValue=record.getValueByPropertyIdentifier( propertyIdentifier=admin.processActionProperty);
+					}else if(structKeyExists(arguments.rc, 'recordAlias')){
+						var processActionProperty=arguments.rc.recordAlias & 'ID';
```
rmarchandTen24: @gerickson808 not all primaryIDs are entity+'ID' example  stateCode on state.cfc. That is why we have getPrimaryIDPropertyName so this might not always work. Is this a smartlist recordAlias vs a collection Alias?

gerickson808: This is a listing display record alias, it's added on as an attribute. It's handled exactly like this (recordAlias & 'ID') when the listing display is initially rendered, however with pagination the whole listing display is rebuilt via a completely different control flow that misses details like this.

PR #5812, opened 2017-10-11
Added sku image to the pricing modal and added image check to the

christopher-kundrat:
thumbnail directive

---
This change is [Reviewable](https://reviewable.io/reviews/ten24/slatwall/5812)


admin/client/src/sku/components/swskuthumbnail.ts
```diff
@@ -9,13 +9,28 @@ class SWSkuThumbnailController{
...
         if(!angular.isDefined(this.skuData)){
             throw("You must provide a sku to the SWSkuThumbnailController");
         }
-        if(angular.isDefined(this.skuData.imagePath)){
-            this.image = this.skuData.imagePath;
-        }
+
+        fileService.imageExists(this.skuData.imagePath).then(
+            ()=>{
+                this.skuData.imagePath = this.skuData.imagePath;
```
jasperboyd: @christopher-kundrat I'm not sure why this would be necessary

christopher-kundrat: @jasperboyd - so at this point we haven't verified if the image from swaddskupricemodallauncher.ts actually exists so it's just running a check and setting it to the placeholder if not found.

jasperboyd: @christopher-kundrat I'm specifically just talking about this line `this.skuData.imagePath = this.skuData.imagePath;`

rmarchandTen24: agreed it would not do anything

christopher-kundrat: yeah that's left over from when the image path was being passed in through a different variable. I'll remove it when I resolve the conflict

PR #5832, opened 2017-10-18
Feature ck rules engine entity build out

christopher-kundrat:


jasperboyd:
@christopher-kundrat sweet the diff just looked crazy

PR #5842, opened 2017-10-23
Develop ck update site relationships

christopher-kundrat:


model/service/OrderService.cfc
```diff
@@ -937,6 +937,11 @@ component extends="HibachiService" persistent="false" accessors="true" output="f
 		//Setup Site Origin if using slatwall cms
 		if(!isNull(getHibachiScope().getSite()) && getHibachiScope().getSite().isSlatwallCMS()){
 			arguments.order.setOrderCreatedSite(getHibachiScope().getSite());
+		}else if (structKeyExists(arguments.data, 'orderCreatedSite')){
+			var createdSite = getService('SiteService').getSite(arguments.data.orderCreatedSite);
```
rmarchandTen24: I think this might be off. the data section comes from the form and will only contain simple data. The process object should be handling inflating simple data to objects. At that point you won't need to do a site service look up because the process object will already have the site

christopher-kundrat: @rmarchandTen24 - good catch. Just pushed an update for that.

PR #5856, opened 2017-11-02
RemoteID additions for Sotheby's import, collection keyword option change

gerickson808:


admin/views/main/log.cfm
```diff
@@ -58,7 +58,7 @@ Notes:
 	</cfif>
 <cfelseif structKeyExists(server, "lucee")>
 	<!--- for clean lucee install --->
-	<cfset local.logFile = expandPath('/') & "WEB-INF/lucee/logs/Slatwall.log">
+	<cfset local.logFile = "/opt/lucee/web/logs/Slatwall.log">
```
sumitverma: This is not standard lucee installation. Probably better to make that a setting to see if there a way to get lucee log location from server scope.

PR #5862, opened 2017-11-06
Develop ck add site inheritance

christopher-kundrat:


rmarchandTen24:
@tdhatcher did you get a chance to look at this. Just want to make sure this doesn't conflict with what you are already doing.

tdhatcher:
The account lookup in the setting service shouldn't be necessary with new site setting functionality. It will check the site associated with the hibachiScope.getCurrentRequestSite()

christopher-kundrat:
The only concern with using the hibachiScope.getCurrentRequestSite() is that we won't have a request site in the Slatwall Admin and there's a need to run site based logic while in there.

PR #5867, opened 2017-11-08
Update Detail Order changes in regards to new location settings logic

christopher-kundrat:
added child lookups to the stock getQATS

### DIFF
--- a/admin/views/entity/ordertabs/addstock.cfm
+++ b/admin/views/entity/ordertabs/addstock.cfm
@@ -57,7 +57,7 @@ Notes:
 
 <!--- Setup default stock location filter--->
 <cfif !isnull(rc.order.getDefaultStockLocation())>
-	<cfset local.addOrderItemStockOptionsSmartList.addFilter("location.locationName", "#rc.order.getDefaultStockLocation().getLocationName()#")>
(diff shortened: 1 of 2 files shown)


admin/views/entity/ordertabs/addstock.cfm
```diff
@@ -75,7 +80,7 @@ Notes:
 		<hb:HibachiListingColumn tdclass="primary" propertyIdentifier="sku.product.productName" />
 		<hb:HibachiListingColumn propertyIdentifier="sku.product.productType.productTypeName" />
 		<hb:HibachiListingColumn propertyIdentifier="sku.skuDefinition" />
-		<hb:HibachiListingColumn propertyIdentifier="calculatedQATS" />
+		<hb:HibachiListingColumn propertyIdentifier="QATS" />
```
sumitverma: we can't use QATS here. this will be major performance issue because it will calculate for each item in the list. Also, we will be migrating these to collection listing and non persistent is not allowed in collection listing.

christopher-kundrat: @sumitverma - so the issue with that though is that the calculatedQATS of parent locations doesn't get updated along with the child locations. For this though we could update this to only grab leaf stock items and then just add a filter to only grab the leafs related to the default location.

sumitverma:
@rmarchandTen24 can you please review and merge.

PR #5868, opened 2017-11-09
Site-level settings based on request site

tdhatcher:
Added logic to apply site-level settings when request is associated with a site. This does create the possibility to reduce clarity for admin users as to which settings are active in all site contexts. Because they may not be actively applied and obvious within the admin experience an admin user could overlook that a site setting is active. Just due to the contextual nature of "sites". 

> In a way the Slatwall admin could be viewed as a "site" as well.


tdhatcher:
@sumitverma @gerickson808 @rmarchandTen24 

Here is an example using the Sku entity and the hierarchy of how the setting value is resolved. It can get quite complex to find a configuration issue when settings are not carefully placed.

1. sku & site
2. sku
3. product & site
4. product 
5. brand & site
6. brand
7. productType & site
8. productType
9. site
10. global
11. settingMetdata.defaultValue

PR #5888, opened 2017-11-13
Added locationCode property

gerickson808:


gerickson808:
@rmarchandTen24 Fixed unique constraint

PR #5905, opened 2017-11-15
Develop kk sku minmax reporting

chriskent:
Removed duplicate processMinMaxSetup_updateStockMinMax in StockService


rmarchandTen24:
Hi Chris can you remove the custom configs. Also a simple rep is required if you don't have a property of objectname

PR #5913, opened 2017-11-15
Develop ck fedex

christopher-kundrat:


integrationServices/fedex/Integration.cfc
```diff
@@ -74,7 +74,8 @@ component accessors="true" output="false" extends="Slatwall.integrationServices.
 			shipperCountryCode = {fieldType="text"},
 			contactPersonName = {fieldType="text"},
 			contactCompany = {fieldType="text"},
-			contactPhoneNumber = {fieldType="text"}
+			contactPhoneNumber = {fieldType="text"},
+			labelImageType = {fieldType="text", defaultValue="PDF"}
```
rmarchandTen24: we might want to make the field type a select and then you can safely pick the options

christopher-kundrat: @rmarchandTen24 - agreed. Just updated this to use the select and included all the allowed types by FedEx

PR #5917, opened 2017-11-21
Updates the actioncaller button

ian-a-hickey:
Adding a check on keypress 'Enter' and submits the form on enter press. This should make it so every form that uses the action caller button can be submitted using the enter key.


rmarchandTen24:
@ian-a-hickey if we have multiple actioncallers which could happen on a detail page in the future. Then we risk submitting multiple forms or run into a race condition where a form is not submitted because it was hijacked by a resolved form submit from a different action caller before it could submit it.

rmarchandTen24:
@ian-a-hickey is action caller already associated with a specific form?

PR #5925, opened 2017-11-23
Add test cases for giftCardService

prasadvishal:


custom/config/configApplication.cfm
```diff
@@ -0,0 +1,2 @@
+<cfset this.name = "slatwall-master">
```
rmarchandTen24: @prasadvishal any code inside the custom folder should be excluded from the pr. If you go to your .git/info folder and add an exclude file where the first line says /custom, you should be able to prevent git from commiting to this folder entirely. Its like a gitignore but personal to you rather than shared. We do this because we want to be able to still commit for custom projects

prasadvishal:
Will this work for now? 
We have excluded the /custom confg files.

rmarchandTen24:
@prasadvishal looks good thanks vishal

PR #5945, opened 2017-12-05
Adding Export to the the Simple Listing Display Controls

jasperboyd:


admin/views/entity/listproduct.cfm
```diff
@@ -99,14 +99,14 @@ Notes:
...
 		<sw-listing-column data-property-identifier="productID" data-is-visible="false" data-is-deletable="false" ></sw-listing-column>
-		<sw-listing-column data-property-identifier="productType.productTypeName" ></sw-listing-column>
-		<sw-listing-column data-property-identifier="brand.brandName" ></sw-listing-column>
-		<sw-listing-column data-property-identifier="productName" tdclass="primary" ></sw-listing-column>
-		<sw-listing-column data-property-identifier="productCode" ></sw-listing-column>
-		<sw-listing-column data-property-identifier="defaultSku.price" ></sw-listing-column>
-		<sw-listing-column data-property-identifier="activeFlag" ></sw-listing-column>
-		<sw-listing-column data-property-identifier="publishedFlag" ></sw-listing-column>
-		<sw-listing-column data-property-identifier="calculatedQATS" ></sw-listing-column>
+		<sw-listing-column data-property-identifier="productType.productTypeName" data-is-exportable="true"></sw-listing-column>
```
rmarchandTen24: @sumitverma @jasperboyd do we need to add it explicitly to productlist? Maybe we just default columns to exportable?

jasperboyd: @rmarchandTen24 absolutely wasn't sure where to change it to make that happen, it seems like this is getting set as false by default in `sw-listing-column` when not passed in, so the backend is just trusting the collection config from the client side

PR #5947, opened 2017-12-06
Add conditon to prevent dublicate email id when guest user becomes pe…

anant-dev:
…rmanent user

updated code for "make sure validation prevents a non-authenticated user than is converted to authenticated user does not use an existing authenticated user email " task.


rmarchandTen24:
@anant-dev can we instead make a validation in the .json. Validations are ideally localized to one spot

anant-dev:
@rmarchandTen24 The duplicate email problem was resolved using separate branch so in order to reflect it in hotfix-mindfire I had to commit it again. Do review and provide your valuable update .
Thanks

PR #5955, opened 2017-12-07
Adds the redeemed codes into the view under the promotion code tab.

ian-a-hickey:
This view was in a custom project. Sumit suggested we could use it in core so this is the pull request.


admin/views/entity/promotiontabs/promotioncodesredeemed.cfm
```diff
@@ -0,0 +1,100 @@
...
+
+--->
+<cfimport prefix="swa" taglib="../../../../tags" />
+<cfimport prefix="hb" taglib="../../../../org/Hibachi/HibachiTags" />
+
+<cfparam name="rc.promotion" type="any">
+<cfparam name="rc.edit" type="boolean">
+
+
+<cfquery name="getRedeemedPromoCodes">
```
rmarchandTen24: @ian-a-hickey can you put this into a dao method?

PR #5957, opened 2017-12-07
deploy cms skeleton site based on setting not by default

jasperboyd:


rmarchandTen24:
@jasperboyd i think the way this currently should work is that it deploys only what doesn't exist which makes sense. What was happening during the update? Was it replacing?

jasperboyd:
@rmarchandTen24 the problem is with the app level inheritance, in the case were a template is being inherited from the app the template doesn't exist at the site level. The update is copying the file back in because it's not there which we definitely don't want in that scenario. That's why I did it as a setting

PR #5967, opened 2017-12-08
Develop pj add email and address type to vendor address

pjohn1024:


rmarchandTen24:
@pjohn1024 @sumitverma this looks like an opportunity to get email address into the slatwalladdress tag. I mean it seems reasonable to me that if the address.cfc has emailaddress the tag should too.

pjohn1024:
@rmarchandTen24 Sure, I can make that change @sumitverma Please let us know what you think.

sumitverma:
Sure, makes sense to me.

pjohn1024:
@rmarchandTen24 My bad...Email address was already in the tag. So cleaned it up.

PR #5979, opened 2017-12-11
able to get lifetime value by currency

rmarchandTen24:


rmarchandTen24:
this should be by orderpayment

PR #5989, opened 2017-12-13
Added an action that calls updateService's function to migrate attributes to custom properties

evan1024:


model/service/AttributeService.cfc
```diff
@@ -186,23 +186,40 @@ component  extends="HibachiService" accessors="true" {
...
+
+		var entityName = arguments.attribute.getAttributeSet().getAttributeSetObject();
+		var customPropertyName = arguments.attribute.getAttributeCode();
+
+		if(this.getEntityHasPropertyByEntityName(entityName,customPropertyName)){
+			getService('UpdateService').migrateAttributeToCustomProperty(entityName,customPropertyName,processObject.getOverrideData());
+			return arguments.attribute;
+		} else {
+			arguments.attribute.addError('Custom Property Does Not Exist','This custom property does not exist in the entity.');
+			return arguments.attribute;
```
migueltarga: if you return outside of the if/else, you only need to return once instead of two times.

evan1024: Fixed.

migueltarga: git push, please haha

rmarchandTen24:
@evan1024 can we make this into a method validation https://github.com/ten24/slatwall/pull/5989/files?w=1#diff-7c8f3641720c438465adc3793d075d15R198

evan1024:
@rmarchandTen24 moved the validation to a validation method

model/service/AttributeService.cfc
```diff
@@ -186,23 +186,36 @@ component  extends="HibachiService" accessors="true" {
 
 	// ===================== START: Process Methods ===========================
 
+	public any function processAttribute_MigrateToCustomProperty(required any attribute, required any processObject, struct data={}) {
+
+		getService("HibachiTagService").cfSetting(requesttimeout="3600");
```
ian-a-hickey: A hour seems like a long time to have the request going if it can't do its job. Maybe some thing like 10 makes more sense?

PR #5996, opened 2017-12-14
Added custom property flag to attributes. On update, the flagged attr…

evan1024:
…ibutes will be copied into the Entity file

### DIFF
--- a/model/service/UpdateService.cfc
+++ b/model/service/UpdateService.cfc
@@ -47,19 +47,19 @@ Notes:
 
 --->
 <cfcomponent extends="HibachiService" accessors="true">
-	
+
 	<!--- prepare constants and environment variables --->
 	<cffunction name="init" returntype="void" >
 		<cfscript>
 			getEnvironmentVariables();
-			
+
 		</cfscript>
 	</cffunction>
-	
+
 	<cffunction name="getEnvironmentVariables" returntype="void">
 		<cfscript>
 			variables.lineBreak = getService('HibachiUtilityService').getLineBreakByEnvironment(getApplicationValue("lineBreakStyle"));
-			
+
 			variables.paddingCount = 2;
 			variables.conditionLineBreak="";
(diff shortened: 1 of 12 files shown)


model/service/UpdateService.cfc
```diff
@@ -370,55 +394,77 @@ Notes:
...
+
 			if(lcase(getApplicationValue("lineBreakStyle")) == 'windows'){
 				conditionalLineBreak = "";
 			}
 			var newContent = "";
 			//add properties
-			if(len(arguments.customEntityParser.getPropertyString())){
+
+			var propertyString = '';
+			if(!isNull(arguments.customEntityParser.getFilePath())){
```
ian-a-hickey: this is checking if customEntityParser.getFilePath() is null, but then using getPropertyString(). The null check needs to be on getPropertyString()?

model/service/UpdateService.cfc
```diff
@@ -370,55 +394,77 @@ Notes:
...
 					;
 					arguments.coreEntityParser.setFileContent(newContent);
-				} 
+				}
 			}
 			//add functions
-			if(len(arguments.customEntityParser.getFunctionString())){
+			var functionString = '';
+			if(!isNull(arguments.customEntityParser.getFilePath())){
+				functionString = arguments.customEntityParser.getFunctionString();
```
ian-a-hickey: Same as above.

model/service/UpdateService.cfc
evan1024: getPropertyString() calls a number of other methods that depend ultimately on the file path being set. During testing, if one tried to import an attribute that had no correspondent custom/entity folder, I noticed that calling propertyString() without first checking if the file path was null caused it to crash. That's why I decided to write it this way.

Another solution might be overriding the getPropertyString() function and returning null if the file path is null, then checking for that. What do you think of that?

PR #5998, opened 2017-12-16
lcase the directory

rmarchandTen24:
This change is [Reviewable](https://reviewable.io/reviews/ten24/slatwall/5998)


org/Hibachi/ckfinder/config.cfm
```diff
@@ -182,15 +182,15 @@ config.resourceType = arrayNew(1);
 config.resourceType[1] = structNew();
 config.resourceType[1].name = 'Files';
 config.resourceType[1].url = config.baseUrl & 'Files';
```
sumitverma: You will need to change URL as well

sumitverma: Also, I think this will be a braking change on case sensitive system, we will need to do it through config.

rmarchandTen24: @sumitverma so are you going to do it via the custom config overrider?

sumitverma: Right, I think we should make these file names variable that you can override from the custom config.

PR #6014, opened 2017-12-19
swlistingdisplay print functionality / supporting collection print on backend

jasperboyd:


org/Hibachi/client/src/listing/components/swlistingdisplay.ts
```diff
@@ -225,7 +227,7 @@ class SWListingDisplayController{
         if (this.collectionObject){
                 this.exampleEntity = this.$hibachi.getEntityExample(this.collectionObject);
         }
-        this.observerService.attach(this.getCollectionByPagination,'swPaginationAction',this.tableID);
```
rmarchandTen24: @migueltarga shouldn't this be using the ID?

migueltarga: yes, everything should have IDs, we can't just shoot messages to the air.

jasperboyd: My bad this came over when I copied the file from the custom project

PR #6019, opened 2017-12-20
Added sorting to return tempaltes

christopher-kundrat:


model/service/ContentService.cfc
```diff
@@ -82,6 +82,7 @@ component extends="HibachiService" persistent="false" accessors="true" output="f
 		if(directoryExists(templateDirectory)) {
 			var directoryList = directoryList(templateDirectory,false,"query","*.cfm|*.html");
```
rmarchandTen24: @christopher-kundrat directoryList supports sorting
https://helpx.adobe.com/coldfusion/cfml-reference/coldfusion-functions/functions-c-d/DirectoryList.html

christopher-kundrat: @rmarchandTen24 - like you mentioned, we need to sort at the end after we add the app templates into the array

PR #6023, opened 2017-12-21
Added auto-incrementing referenceNumber property to stockAdjustment e…

tdhatcher:
…ntity and views.


model/service/StockService.cfc
```diff
@@ -455,6 +455,16 @@ component extends="HibachiService" accessors="true" output="false" {
 		// Set the status to closed
 		arguments.stockAdjustment.setStockAdjustmentStatusType( getTypeService().getTypeBySystemCode("sastClosed") );
 
+		// Set reference number
+		lock scope="Application" timeout="30" {
```
rmarchandTen24: @tdhatcher the locks are not reliable when we have multiple servers. You should look at how orderNumber currently works. It puts the work of incrementing onto sql to ensure no duplicates

PR #6024, opened 2017-12-21
made recordProcessContext always lowercase in hibachi tag

evan1024:


org/Hibachi/HibachiTags/HibachiListingDisplay.cfm
```diff
@@ -246,7 +246,10 @@
 			<!--- Process --->
 			<cfif len(attributes.recordProcessAction) and attributes.recordProcessButtonDisplayFlag>
 				<cfset attributes.administativeCount++ />
-
+				
+				<!-- make sure recordProcessContext is always lower case -->
+				<cfset attributes.recordProcessContext = lCase(attributes.recordProcessContext) />
```
rmarchandTen24: @evan1024 can we just lowercase it in the instance where it is being used on an include or referencing a cfm? Not sure we want to lcase the process context 100% for everything.

evan1024: Fixed.

https://github.com/ten24/slatwall/pull/6024/commits/eee76866b86ba0d8af9aeff66ab01aa938781558

PR #6048, opened 2018-01-04
Develop kk stock minmax reporting

chriskent:
Added product.activeFlag to sku collection
Fixed column headings for from & to location
Removed ability to save collection "usingPersonalCollection="false"


org/Hibachi/HibachiTags/HibachiListingDisplay.cfm
```diff
@@ -80,6 +80,7 @@
 	<cfparam name="attributes.createAction" type="string" default="" />
 	<cfparam name="attributes.createQueryString" type="string" default="" />
 	<cfparam name="attributes.exportAction" type="string" default="" />
+	<cfparam name="attributes.usingPersonalCollection" type="string" default="true" />
```
rmarchandTen24: usingPersonalCollection should be defaulted to false we have it explicitly turned on for the the listxxx.cfms explicitly

org/Hibachi/HibachiTags/HibachiListingDisplay.cfm
```diff
@@ -106,7 +107,7 @@
 				data-has-action-bar="false"
 				data-expandable="#attributes.expandable#"
 			    edit="true"
-			    data-using-personal-collection="true"
+			    data-using-personal-collection="#attributes.usingPersonalCollection#"
```
rmarchandTen24: @chriskent i think this was the missing part

PR #6074, opened 2018-01-15
Fix S3 Upload on lucee4

migueltarga:


org/Hibachi/HibachiTransient.cfc
```diff
@@ -415,22 +415,22 @@ component output="false" accessors="true" persistent="false" extends="HibachiObj
 					// If the directory where this file is going doesn't exists, then create it
 					if(left(uploadDirectory, 5) == 's3://'){
 
+						var uploadData = fileUpload('ram://', currentProperty.name, currentProperty.hb_fileAcceptMIMEType, 'makeUnique' );
```
sumitverma: please change ram:// to getVirtualFileSystemPath()

PR #6103, opened 2018-01-20
Moved reference number assignment to init. fixed DAO method

evan1024:


model/entity/StockAdjustment.cfc
```diff
@@ -80,7 +81,17 @@ component entityname="SlatwallStockAdjustment" table="SwStockAdjustment" persist
 	// Deprecated Properties
 	property name="statusCode" persistent="false";		// Use getStockAdjustmentStatusTypeSystemCode()
 	
+	public any function init(){
+		lock scope="Application" timeout="30" {
```
sumitverma: Instead of init we should do this in preinsert or save override method.

model/entity/StockAdjustment.cfc
```diff
@@ -89,6 +100,13 @@ component entityname="SlatwallStockAdjustment" table="SwStockAdjustment" persist
 		}
 	}
 	
+<<<<<<< Updated upstream
+=======
```
sumitverma: Git merge issue, will need cleanup.

model/entity/StockAdjustment.cfc
```diff
@@ -89,6 +100,13 @@ component entityname="SlatwallStockAdjustment" table="SwStockAdjustment" persist
...
+	public boolean function isNotClosed(){
+		return variables.stockAdjustmentStatusType.getSystemCode() != "sastClosed";
+	}
+	
+>>>>>>> Stashed changes
```
sumitverma: cleanpu needed

model/entity/StockAdjustment.cfc
```diff
@@ -243,7 +249,11 @@ component entityname="SlatwallStockAdjustment" table="SwStockAdjustment" persist
 	// =================== START: ORM Event Hooks  =========================
 	
 	public void function preInsert(){
-		super.preInsert();
+		lock scope="Application" timeout="30" {
+	 		var maxReferenceNumber = getDAO('StockDAO').getStockAdjustmentMaxReferenceNumber()[1];
+	 		variables.ReferenceNumber = maxReferenceNumber + 1;
+ 		}
```
tdhatcher: We should probably wrap this in its own method and refactor closer to how Order entity does similar logic.
https://github.com/ten24/slatwall/blob/develop/model/entity/Order.cfc#L248
https://github.com/ten24/slatwall/blob/develop/model/entity/Order.cfc#L1486

PR #6129, opened 2018-01-24
new calculated properties for Order

evan1024:


config/resourceBundles/en.properties
```diff
@@ -3058,6 +3058,16 @@ entity.order.appliedPromotions
 entity.order.assignedAccount                                                                                   = Assigned Account
 entity.order.billingAddress                                                                                    = Billing Address
 entity.order.calculatedTotal                                                                                   = Total
+entity.order.calculatedSubTotal = Calculated Subtotal
```
rmarchandTen24: @evan1024 Calculated properties do not need the word Calculated explicitly. Please remove

PR #6165, opened 2018-02-12
Exclude orderBy createdDateTime in case of hasDisplayAggregate : true and no other orderBy applied

anant-dev:
PR  regarding - Collection bug fix - when you have a display aggergate and no order by then exclude order by createDatetime from being added , Task on Teamwork , @rmarchandTen24  please review

### DIFF
--- a/model/entity/Collection.cfc
+++ b/model/entity/Collection.cfc
@@ -1786,7 +1786,9 @@ component displayname="Collection" entityname="SlatwallCollection" table="SwColl
 			}else{
 				if(!getExcludeOrderBy()){
 					var orderBy = getDefaultOrderBy();
-					groupByList = listAppend(groupByList,orderBy.propertyIdentifier);
+					if(!getHasDisplayAggregate()){
+						groupByList = listAppend(groupByList,orderBy.propertyIdentifier);
+					}
 				}
 			}
 		}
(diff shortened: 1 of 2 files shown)


model/entity/Collection.cfc
```diff
@@ -2835,7 +2843,9 @@ component displayname="Collection" entityname="SlatwallCollection" table="SwColl
 				}
 
 				if(!structKeyExists(collectionConfig,'orderBy') || !arrayLen(collectionConfig.orderBy)){
-					arrayAppend(groupBys,'_' & lcase(getService('hibachiService').getProperlyCasedShortEntityName(getCollectionObject())) & '.' & "createdDateTime");
+					if(!getHasDisplayAggregate()){
+						arrayAppend(groupBys,'_' & lcase(getService('hibachiService').getProperlyCasedShortEntityName(getCollectionObject())) & '.' & "createdDateTime");
```
rmarchandTen24: @anant-dev can you change this line to use the default order by vs createdDateTime?

rmarchandTen24: getDefaultOrderBy()

rmarchandTen24: @migueltarga I think if the default orderby is not createdDatetime then we get a mismatch on the group by

migueltarga: @rmarchandTen24 you are right about that! Good catch

anant-dev: @rmarchandTen24  yes you are right, am replacing it by getDefaultOrderBy().propertyIdentifier

anant-dev:
@rmarchandTen24 made changes as per the suggestion in the review. Thanks for pointing it out.

anant-dev:
@rmarchandTen24  please review the PR and the **latest commit adding tests to check if orderBy createdDateTime is excluded in case of hasDisplayAggregate is true when orderBy applied or not applied**

PR #6176, opened 2018-02-19
Work around for quantity based pricing data getting incorrectly set to

christopher-kundrat:
null

### DIFF
--- a/admin/client/src/sku/components/swaddskupricemodallauncher.ts
+++ b/admin/client/src/sku/components/swaddskupricemodallauncher.ts
@@ -51,6 +51,7 @@ class SWAddSkuPriceModalLauncherController{
         //these are populated in the link function initially
         this.skuPrice = this.entityService.newEntity('SkuPrice'); 
         this.skuPrice.$$setSku(this.sku);
+        
         if(angular.isUndefined(this.disableAllFieldsButPrice)){
             this.disableAllFieldsButPrice = false; 
         }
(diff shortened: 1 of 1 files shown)


admin/client/src/sku/components/swaddskupricemodallauncher.ts
```diff
@@ -126,13 +131,15 @@ class SWAddSkuPriceModalLauncherController{
...
                 this.saveSuccess = false; 
             }
-        ).finally(()=>{
+        ).finally(()=>{       
             if(this.saveSuccess){
                 
                 for(var key in this.skuPrice.data){
-                    this.skuPrice.data[key] = null;
+                    if (key != 'sku' && key !='currencyCode'){
+                        this.skuPrice.data[key] = null;
```
jasperboyd: @christopher-kundrat this doesn't really make sense to me, sku and currency get set in the `initData` function in this file that gets called on line 144 https://github.com/ten24/slatwall/blob/993d91df5cddf5ffd106d2ab498c76a43e118801/admin/client/src/sku/components/swaddskupricemodallauncher.ts#L50-L78

What's the issue this is trying to solve?

christopher-kundrat: @jasperboyd - after submitting the form you aren't able to use it again without reloading the page because the CurrencyCode and the selected sku are getting wiped. Long term solution would be to fix the resetForm functionality but this was a quick workaround to get it working.

jasperboyd: @christopher-kundrat that's what I'm saying I don't see how they're getting wiped they're getting set back on the model in the code I reference above

PR #6182, opened 2018-02-26
Add condition and tests for cached and new collection list

anant-dev:
@rmarchandTen24 please review the PR regarding cached and new collection list conditions added in HibachiEntity and test cases to verify it in the collectiontest.cfc


anant-dev:
@rmarchandTen24 please review the changes

PR #6207, opened 2018-03-16
Fix & Migrate WorkflowTaskConfig

migueltarga:


rmarchandTen24:
@migueltarga you need to have the code that fixes the alias going forward to compliment the migration

PR #6219, opened 2018-03-23
 Update TestBox Directory Mapping UI with drop downs

anant-dev:
Add UI drop downs separately for test folder, file, and method in TestBox (test-runner) UI.

### DIFF
--- a/meta/tests/test-runner/directoryMapping.cfc
+++ b/meta/tests/test-runner/directoryMapping.cfc
@@ -1,0 +1,46 @@
+/**
+ * dirStruct
+ *
+ * @author anants
+ * @date 3/19/18
+ **/
+component accessors=true output=false persistent=false {
+
+	remote any function getTestFolders() returnFormat="json"{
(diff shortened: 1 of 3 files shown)


meta/tests/test-runner/directoryMapping.cfc
```diff
@@ -0,0 +1,46 @@
+/**
+ * dirStruct
+ *
+ * @author anants
+ * @date 3/19/18
+ **/
+component accessors=true output=false persistent=false {
+
+	remote any function getTestFolders() returnFormat="json"{
+	path ="/Slatwall/meta/tests/unit/";
```
rmarchandTen24: @anant-dev  none of this appears to be var scoped. If it is a part of a component it should be var scoped

anant-dev: ok @rmarchandTen24 will include var scope

rmarchandTen24:
@anant-dev This looks good. Can I get a pr with this work directly into Hotfix. I'd love to start using this!

PR #6226, opened 2018-03-26
Develop ar totalandavgrows

anthonyr-ten24:


org/Hibachi/client/src/listing/components/listingdisplay.html
```diff
@@ -106,13 +106,13 @@ <h2 data-ng-bind="swListingDisplay.title"></h2>
                                     <label for="exampleID"></label>
                                 </div>
                             </th>
-                            <!-- TODO: CREATE FLAG IN COLUMN TO SET CLASS 's-detail-hidden-content'-->
+                            <!-- TODO: CREATE FLAG IN COLUMN TO SET CLASS 's-detail-hidden-content'--> 
 
-                            <th colspan="2"  data-ng-click="column.persistent.toString() != 'false' && swListingDisplay.toggleOrderBy(column)" data-ng-repeat="column in swListingDisplay.columns" class="data {{column.tdClass}}" data-ng-class="{'s-detail-hidden-content': column.CHANGE_THIS_LATER}" data-ng-if="column.isVisible" ng-cloak>
+                            <th colspan="2" data-ng-repeat="column in swListingDisplay.columns" class="data {{column.tdClass}}" data-ng-class="{'s-detail-hidden-content': column.CHANGE_THIS_LATER}" data-ng-if="column.isVisible" ng-cloak>
```
rmarchandTen24: @anthonyr-ten24 can we remove the colspan="2" stuff?

rmarchandTen24: @anthonyr-ten24 I think we don't need colspan=2 anymore because we aren't doing the rows the way that needed it before

anthonyr-ten24: If we remove it, this will happen 
screen shot 2018-03-26 at 1 32 39 pm

rmarchandTen24: @anthonyr-ten24  i think the colspan needs to be removed in other places as well. I'm very sure of this because I added it

PR #6243, opened 2018-03-30
Add custom aggregrate filter to filter ormtype as numeric and filedtype as id

anant-dev:


anant-dev:
@rmarchandTen24  please review the PR

PR #6268, opened 2018-04-11
adding ability to set cache expiration date time

jasperboyd:


sumitverma:
what will be a good use case for this? Where are you planning to set the expiration date?

jasperboyd:
@sumitverma My use case for adding this was the need to cache a daily report because of it's dependence on the results of multiple collections, In that case I'm caching for 6 hours so users who check at the beginning of the day will see updated values at the end of the day. 

Mostly just give developers some extra flexibility when caching

PR #6274, opened 2018-04-13
Fixed typeahead on inner order pages

anthonyr-ten24:
https://ten24.teamwork.com/#tasks/18055914


rmarchandTen24:
@christopher-kundrat @ian-a-hickey Does this look correct guys?

evan1024:
@rmarchandTen24 Jumping in because I created that attribute; It is optional and it was breaking the typeahead because that are no children locations in the current test DB, causing the typeahead to always return blank.

PR #6277, opened 2018-04-16
added product type validation

desiraefreedman:
added validation for product type names to be unique
added validation for no use of non alphanumeric characters and no spaces for urlTitle field.

### DIFF
--- a/model/validation/ProductType.json
+++ b/model/validation/ProductType.json
@@ -1,7 +1,7 @@
 {
 	"properties":{
-		"productTypeName":		[{"contexts":"save","required":true}],
-		"urlTitle":				[{"contexts":"save","required":true,"unique":true}],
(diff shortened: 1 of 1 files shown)


model/validation/ProductType.json
```diff
@@ -1,7 +1,7 @@
 {
 	"properties":{
-		"productTypeName":		[{"contexts":"save","required":true}],
-		"urlTitle":				[{"contexts":"save","required":true,"unique":true}],
+		"productTypeName":		[{"contexts":"save","required":true,"unique":true}],
+		"urlTitle":				[{"contexts":"save","required":true,"unique":true,"regex":"^[a-zA-Z0-9-_.|:~^]+$"}],
```
migueltarga: See!! We use regex!!! 👍

PR #6279, opened 2018-04-16
validating order delivery quantities

evan1024:


model/process/OrderDelivery_Create.cfc
```diff
@@ -85,17 +86,29 @@ component output="false" accessors="true" extends="HibachiProcess" {
...
 		){
 			variables.shippingIntegration = getOrderFulfillment().getShippingMethodRate().getShippingIntegration();
-		} 
-		if(structKeyExists(variables,'shippingIntegration')){
-			return variables.shippingIntegration;
+		} else {
+			variables.shippingIntegration = "";
 		}
-		
+		return variables.shippingIntegration;
```
rmarchandTen24: @evan1024  this is wrong because we need to be able to support returning a null

evan1024:
@rmarchandTen24 reverted. That wasn't intended to be part of this PR. My bad.

rmarchandTen24:
@evan1024 was there some place that was expecting a string instead of a null?

PR #6290, opened 2018-04-20
Develop team th aws sns email parser

tdhatcher:
Miguel, here is some of the progress with abstracting the general AWS SNS handling logic into Hibachi/Slatwall. I'm still adding the parsing/handling pieces that I've developed first as external units since this functionality is more challenging to develop/test running within the Hibachi/Slatwall lifecycle. So I'm in the process of wiring it all up within the Hibachi/Slatwall lifecycle.

But basically once it has the SNS message deserialized and parsed into a struct (and fetching any corresponding S3 file if needed), an event is announced with the data. That way any future application with other SNS needs can leverage this boilerplate.

A custom event handler will take care of pushing the SNS data into a custom integrations services.

### DIFF
--- a/admin/controllers/aws.cfc
+++ b/admin/controllers/aws.cfc
@@ -1,0 +1,96 @@
+
+/*
+
+    Slatwall - An Open Source eCommerce Platform
+    Copyright (C) ten24, LLC
+
+    This program is free software: you can redistribute it and/or modify
+    it under the terms of the GNU General Public License as published by
+    the Free Software Foundation, either version 3 of the License, or
+    (at your option) any later version.
+
+    This program is distributed in the hope that it will be useful,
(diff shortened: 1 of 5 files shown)


tdhatcher:
@migueltarga 

This is in a stable state in Slatwall core. All a application needs to do is implement a custom event handler with 'onAwsSnsReceive' and then it can broker the data to proper services/entities in the custom application. 

By the time the request reaches the event handler, the SNS payload is all parsed and available as a struct and an associated s3 object (if it exists) has already been fetched and stored in a byte array for further file operations (eg. running through the email parser).

So email messages and attachments are ready to access with less 10 lines of code in a application's custom event handler.

SNS http endpoints will auto subscribe when AWS sends the request so no manual configuration required.

PR #6302, opened 2018-04-24
made changes in OrderDAOTest

mfsi-nirmitm:
made changes in OrderDAOTest unit test cases to solve the failure case

### DIFF
--- a/meta/tests/unit/entity/VendorOrderTest.cfc
+++ b/meta/tests/unit/entity/VendorOrderTest.cfc
@@ -55,6 +55,9 @@ component extends="Slatwall.meta.tests.unit.entity.SlatwallEntityTestBase" {
 		variables.entity = request.slatwallScope.getService("vendorOrderService").newVendorOrder();
 	}
 
+	/**
+	* @test
+	*/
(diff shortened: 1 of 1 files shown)


meta/tests/unit/dao/OrderDAOTest.cfc
```diff
@@ -77,7 +77,7 @@ component extends="Slatwall.meta.tests.unit.dao.SlatwallDAOTestBase" {
 				{
 					orderPaymentID='',
 					orderPaymentStatusType={
-						orderPaymentStatusTypeID = '5accbf57dcf5bb3eb71614febe83a31d'
+						orderPaymentStatusTypeID = '444df31fa8adde8d71c5ca279e42a00d'
```
rmarchandTen24: @mfsi-nirmitm This is an id for an unrelated type. This is a subscription status type not a an orderPaymentStatusType. Why would you use this id?

mfsi-nirmitm: Okay , i got the point i have updated the pull. 
Thanks.

PR #6344, opened 2018-05-01
shopping cart template: dynamic cart items section now pulls data fro…

evan1024:
…m slatwall and updates it through an attribute directive


org/Hibachi/client/src/frontend/components/swfcartitems.ts
```diff
@@ -0,0 +1,64 @@
...
+        return directive;
+    }
+    
+    //@ngInject
+    constructor(
+        $rootScope
+    ){
+        return {
+            restrict: "A",
+            require: "^ngModel",
```
rmarchandTen24: @evan1024 instead of requiring ngModel we should have a binding on scope

integrationServices/slatwallcms/skeletonsite/templates/inc/cartItems.cfm
```diff
@@ -0,0 +1,98 @@
+<div ng-if="false" class="alert alert-success">Item removed from cart</div>
+<div ng-if="false" class="alert alert-danger">Item removed failure</div>
+<div ng-if="false" class="alert alert-success">Quantity updated</div>
+<div ng-if="false" class="alert alert-danger">Quantity update failure</div>
+
+<div class="row" ng-repeat-start="orderItem in slatwall.cart.orderItems" swf-cart-items ng-model="orderItem">
```
rmarchandTen24: ng-model is for <input><select> or user input tags

this shoudl be a binding on the swf-cart-items directive

org/Hibachi/client/src/frontend/components/swfcartitems.ts
```diff
@@ -0,0 +1,64 @@
+/// <reference path='../../../typings/hibachiTypescript.d.ts' />
+/// <reference path='../../../typings/tsd.d.ts' />
+/**
+ * True if the data type matches the given data type.
+ */
+/**
+ * Validates true if the model value is a numeric value.
+ */
+class SWFCartItemsController{
```
rmarchandTen24: @evan1024 we should meet on how to get this to use controllerAs syntax so that upgrading will be easy

PR #6346, opened 2018-05-01
promobox directive

evan1024:


integrationServices/slatwallcms/skeletonsite/templates/inc/promoBox.cfm
```diff
@@ -1,24 +1,24 @@
-<div class="card mb-4">
+<div class="card mb-4" swf-promo-box ng-model="promoBox" ng-cloak>
```
rmarchandTen24: @evan1024 I think this is another example of needing a binding on swf-promo-box rather than ng-model

org/Hibachi/client/src/frontend/components/swfpromobox.ts
```diff
@@ -0,0 +1,55 @@
...
+        return directive;
+    }
+    
+    //@ngInject
+    constructor(
+        $rootScope
+    ){
+        return {
+            restrict: "A",
+            require:"^ngModel",
```
rmarchandTen24: @evan1024 change this to a binding on scope and lets convert this to controllerAs syntax

PR #6355, opened 2018-05-03
cancelling collection requests on the client side if a new one is made within a second

jasperboyd:


org/Hibachi/client/src/collection/controllers/collections.ts
```diff
@@ -96,33 +96,38 @@ class CollectionController{
...
-				}
-				collectionService.setFilterCount(filterItemCounter());
-				$scope.loadingCollection = false;
-			},function(reason){
-			});
-            return collectionListingPromise;
+			if($scope.getCollectionPromise){
+				$timeout.cancel($scope.getCollectionPromise);
+			}
+			$scope.getCollectionPromise = $timeout(function(){
```
gerickson808: It seems like this will wait a full second before making any collection request, no?

jasperboyd: @gerickson808 yep the issue is that when you're updating a bunch of display options setting visible it makes 2 requests everytime you make a change. If you're making a lot of changes really fast you can spam the server with 30-40 requests, this waits for the user to stop before going out and asking for the data

jasperboyd:
no problem @rmarchandTen24

PR #6370, opened 2018-05-08
Updates SlatwallSmartListPager to URLEncodedFormat

ian-a-hickey:
unsafe data that can be passed into a collection.


sumitverma:
Will it make sense to do this in build url method instead?

ian-a-hickey:
Agreed. Closing

PR #6378, opened 2018-05-09
Created new core slatwall email templates 

erinmooneyten24:
These templates are now responsive and a lot prettier. Of course, this means a lot more code, but I added a bunch of comments so it should still be easy to understand.


rmarchandTen24:
@erinmooneyten24 for the slatwall projects we need to exclude committing anything in the custom folder. You can do this by adding a "git exclude". Go to the .git folder at the root of your project, and add an info folder and then inside the info folder add a file named exclude.
then inside the exclude folder add the line "custom". This will exclude that folder from being committed.

PR #6416, opened 2018-05-22
Upgrade Modules to Angular 2 and standardize module files

anant-dev:


admin/client/src/app.module.ts
```diff
@@ -35,21 +46,30 @@ import { parseProvider,logProvider,filterProvider,timeoutProvider,qProvider } fr
...
+    OrderFulfillmentModule,
     OrderItemModule,
     ProductModule,
-    FormModule,
-    Pagination,
-    ValidationModule,
-    WorkflowModule
+    ProductBundleModule,
+    SkuModule,
+    SlatwallAdminModule
```
rmarchandTen24: @anant-dev once all dependencies are settled. I think we should be able to get away with just adding the slatwalladmin module here

anant-dev: @rmarchandTen24  Slatwalladmin module doesn't carry all the modules imported into it, but has a few of them, so should I remove the ones imported in slatwalladmin module.

PR #6424, opened 2018-05-29
Support to Process Context in Action Permission

migueltarga:


org/Hibachi/HibachiAuthenticationService.cfc
```diff
@@ -44,7 +44,10 @@ component output="false" accessors="true" extends="HibachiService" {
 		} else {
 			var itemName = 'default';
 		}
-		
+		if(structKeyExists(arguments, 'processContext') && len(arguments.processContext)){
+			itemName &= '_#arguments.processContext#';
```
christopher-kundrat: With this change we'll just need to updated the entityname lookup farther down to not include the processContext

PR #6450, opened 2018-06-08
Turn Attributes into real properties

migueltarga:


config/scripts/preupdate/v5_1.cfm
```diff
@@ -0,0 +1,23 @@
...
+	<cfset local.found = false />
+	<cfloop query="#attributecolumns#">
+		<cfif Column_Name eq 'customPropertyFlag'>
+			<cfset local.found = true />
+			<cfbreak>
+		</cfif>
+	</cfloop>
+	<cfif !local.found>
+		<cfquery name="local.addCustomPropertyFlag" datasource="#this.datasource.name#">
+			ALTER TABLE SwAttribute ADD customPropertyFlag tinyint(1)
```
sumitverma: Why do we need this script? Property will automatically get created.

migueltarga: We query attributes before ormflush, and this column won't exist

PR #6468, opened 2018-06-12
Sku image upload formatting fixes

anthonyr-ten24:


admin/client/src/sku/components/imagedetailmodallauncher.html
```diff
@@ -25,7 +25,7 @@
                 </div>-->
 
                 <div class="alert alert-info popup-alert-message" id="j-multi-sku-message">
-                    <strong>( {{swImageDetailModalLauncher.skusAffectedCount}} ) skus with the same option value will be effected.</strong>
+                    ({{swImageDetailModalLauncher.skusAffectedCount}}) skus with the same option value will be effected.
```
gerickson808: Should be "affected"

PR #6483, opened 2018-06-15
fix to not recalc order amount total when equal to zero

desiraefreedman:


model/entity/Order.cfc
```diff
@@ -356,15 +356,6 @@ component displayname="Order" entityname="SlatwallOrder" table="SwOrder" persist
 		return arr;
 	}
 
-	public numeric function getPaymentAmountDueAfterGiftCards(){
```
ian-a-hickey: @desiraefreedman Did you mean to remove this method? I think its needed by the gift card processing logic.

PR #6496, opened 2018-06-20
redirect correctly

ian-a-hickey:
Because this was not appending the /Slatwall, it was redirecting to homepage of a custom site.

### DIFF
--- a/admin/client/src/orderfulfillment/components/sworderfulfillmentlist.ts
+++ b/admin/client/src/orderfulfillment/components/sworderfulfillmentlist.ts
@@ -597,9 +597,11 @@ class SWOrderFulfillmentListController {
         if (result.data && result.data['fulfillmentBatchID']){
             //if url contains /Slatwall use that
             var slatwall = "";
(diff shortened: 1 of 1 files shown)


admin/client/src/orderfulfillment/components/sworderfulfillmentlist.ts
```diff
@@ -595,7 +595,12 @@ class SWOrderFulfillmentListController {
         //Redirect to the created fulfillmentBatch.
         this.addingBatch = false;
         if (result.data && result.data['fulfillmentBatchID']){
-            this.$window.location.href = "/?slataction=entity.detailfulfillmentbatch&fulfillmentBatchID=" + result.data['fulfillmentBatchID'];
+            //if url contains /Slatwall use that
+            var slatwall = "";
+            if (this.$window.location.pathname.indexOf("Slatwall") != -1){
```
rmarchandTen24: @ian-a-hickey I think we should be using the baseurl variable in appconfig

PR #6497, opened 2018-06-20
remove shipping charge from shipping method options if handling fee exists

DenivalHeitz:


model/service/OrderService.cfc
```diff
@@ -3545,7 +3545,11 @@ component extends="HibachiService" persistent="false" accessors="true" output="f
 		for(var shippingMethodOption in arguments.orderFulfillment.getFulfillmentShippingMethodOptions()) {
 
 			var thisOption = {};
-			thisOption['name'] = shippingMethodOption.getSimpleRepresentation();
+			if(arguments.orderFulfillment.hasHandlingFeeApplied()){
+				thisOption['name'] = shippingMethodOption.getShippingMethodRate().getShippingMethod().getShippingMethodName()
```
migueltarga: missing semi column

rmarchandTen24: @DenivalHeitz where does hasHandlingFeeApplied come from? OrderFulfillment does not have this method.

PR #6507, opened 2018-06-25
Added functionality to update subscription status on

christopher-kundrat:
renewal errors

### DIFF
--- a/model/dao/SubscriptionDAO.cfc
+++ b/model/dao/SubscriptionDAO.cfc
@@ -175,5 +175,33 @@ Notes:
 
 		<cfreturn ormExecuteQuery(hql, {productID=arguments.productID}) />
 	</cffunction>
+	
+	<!--- This is so that we can update subsction statues when there are errors on the order --->
+	<cffunction name="updateSubscriptionStatus">
+		<cfargument name="subscriptionUsageID" required="true" type="string" />
+		
+		<cfset var subscriptionStatusID = replace(lcase(createUUID()), '-', '', 'all') />
+		<cfset var currentDate = now() />
+		
+		<cfquery name="createNewSubscriptionStatus">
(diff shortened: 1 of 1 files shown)


model/service/SubscriptionService.cfc
```diff
@@ -604,6 +604,14 @@ component extends="HibachiService" persistent="false" accessors="true" output="f
 			
 			arguments.subscriptionUsage.addError('renew', errorResponseMessage);
 		}
+		
+		if (order.hasErrors() ||  arguments.subscriptionUsage.hasErrors() ) {
+			
+			var currentStatus = arguments.subscriptionUsage.getCurrentStatus();
+			if(currentStatus.getSubscriptionStatusType().getSystemCode() == 'sstActive' && arguments.subscriptionUsage.getExpirationDate() <= now()) {
+				getDAO('SubscriptionDAO').updateSubscriptionStatus(arguments.subscriptionUsage.getSubscriptionUsageID());
```
sumitverma: Why do we need to use DAO, instead of this.processSubscriptionUsage(arguments.subscriptionUsage, {}, 'updateStatus');?

In fact, why do we even need to check any condition and not just call this.processSubscriptionUsage(arguments.subscriptionUsage, {}, 'updateStatus');

sumitverma: You can remove the error from subscription usage, call update and then set it back.

christopher-kundrat: @sumitverma - won't the orm flush still be an issue because of the order and order payment?

sumitverma: You can clear those as well. Main issue I see is having subscription status update logic in multiple places. It should all be consolidated to this.processSubscriptionUsage(arguments.subscriptionUsage, {}, 'updateStatus'). Also, running DAO, skips generating audit records.

christopher-kundrat: Yeah looking through it, it doesn't seem like we need the errors on the order or order payment once they've been added to the subscription usage so that should be fine. I can go ahead and make that update

PR #6518, opened 2018-06-28
Fixing weird issue in buildFormCollections with Lucee

tdhatcher:
Fixing weird issue in buildFormCollections with Lucee auto-creating various form scope variables for keys with array syntax. It will cause buildFormCollections logic to overwrite and change key names back to array syntax causing populate to not work properly. Unintended behavior depends on the arbitrary order the keys arrange themselves in the form scope.

Compare the difference between CF and Lucee with the same inputs.

[CF11 Form Scope.html.pdf](https://github.com/ten24/slatwall/files/2143366/CF11.Form.Scope.html.pdf)

[Lucee5 Form Scope.html.pdf](https://github.com/ten24/slatwall/files/2143367/Lucee5.Form.Scope.html.pdf)

This shows the structure that buildFormCollections creates as it goes over the form scope. One depicts how the variables get overwritten or added to the wrong places. Other shows after the fix.

[Issue - Lucee Form Scope Example.html.pdf](https://github.com/ten24/slatwall/files/2143368/Issue.-.Lucee.Form.Scope.Example.html.pdf)

[Fixed - Lucee Form Scope Example.html.pdf](https://github.com/ten24/slatwall/files/2143369/Fixed.-.Lucee.Form.Scope.Example.html.pdf)

### DIFF
--- a/org/Hibachi/HibachiUtilityService.cfc
+++ b/org/Hibachi/HibachiUtilityService.cfc
@@ -1172,6 +1172,9 @@
 			<!--- If the field has a dot or a bracket... --->
 			<cfif hasFormCollectionSyntax(local.thisField)>
 
+				<!--- NOTE: Edge case issue caused by Lucee automatically converting keys with array syntax and inflating new form scope variables to represent arrays (a Lucee hidden gem) --->
+				<cfset local.isLuceeTerminalCollectionFieldFlag = right(local.thisField, 1) eq ']' />
+
 				<!--- Add collection to list if not present. --->
 				<cfset local.tempStruct['formCollectionsList'] = addCollectionNameToCollectionList(local.tempStruct['formCollectionsList'], local.thisField) />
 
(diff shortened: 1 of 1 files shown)


tdhatcher:
I have tested this on CF11, Lucee 4 & 5.

sumitverma:
There might be some application setting to turn this behavior off and get string instead. Can you please also post to Lucee group to confirm?

tdhatcher:
@sumitverma I also thought about that possibility but did not find any obvious setting. What I did find is application settings name called 'sameFormFieldAsAarray' and 'sameUrlFieldsAsArray'. But they were both false during my tests which is the default. If this is the intent of the setting, there must be a bug then.

And the other mention I found about the form scope's ability to handle array bracket notation behavior was here: [Lucee Hidden Gems](http://docs.lucee.org/guides/cookbooks/Hidden_gems.html#example-3-passing-arguments-via-urlform-scopes-as-array)

On the Hidden Gems page it only mentions the fieldnames country[] with no index or subsequent dot notation. 

Maybe Lucee can change to ignore the [n] and only do so when strictly [], or create/fix some application setting to turn this off.

I will post there.

tdhatcher:
@sumitverma @rmarchandTen24 I've posted to Lucee Dev forum.

https://dev.lucee.org/t/bug-issue-with-form-scope-array-bracket-notation-setting-to-disable/4088

tdhatcher:
@sumitverma @rmarchandTen24 
It looks like at the present moment the Lucee specific handling is needed to get around the undesired behavior.

https://dev.lucee.org/t/bug-issue-with-form-scope-array-bracket-notation-setting-to-disable/4088/4

sumitverma:
ok, thanks.

Sumit Verma

PR #6559, opened 2018-07-18
Develop rm getrequest

rmarchandTen24:
This change is [Reviewable](https://reviewable.io/reviews/ten24/slatwall/6559)

### DIFF
--- a/org/Hibachi/HibachiControllerREST.cfc
+++ b/org/Hibachi/HibachiControllerREST.cfc
@@ -53,11 +53,10 @@ component output="false" accessors="true" extends="HibachiController" {
 
         getFW().setView("public:main.blank");
         arguments.rc.headers["Content-Type"] = "application/json";
-        
+        var functionName = getFW().getItem();
         if(
-            structKeyExists(arguments.rc,getApplicationValue('action')) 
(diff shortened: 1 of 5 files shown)


org/Hibachi/HibachiControllerREST.cfc
```diff
@@ -53,6 +53,13 @@ component output="false" accessors="true" extends="HibachiController" {
 
         getFW().setView("public:main.blank");
         arguments.rc.headers["Content-Type"] = "application/json";
+        var functionName = getFW().getItem();
+        if(
+            len(functionName) >= 3 
+            && left(functionName,3)=='get'
+        ){
+            arguments.rc.context = 'GET';
```
jasperboyd: @rmarchandTen24 do we want to set to POST in the else?

rmarchandTen24: @jasperboyd context on a post is usually a process method so we can't

jasperboyd: that wouldn't be `rc.processContext`?

rmarchandTen24: https://github.com/ten24/slatwall/blob/develop/org/Hibachi/HibachiControllerREST.cfc#L753

PR #6603, opened 2018-08-07
Removed extra commas from type xml

christopher-kundrat:
This change is [Reviewable](https://reviewable.io/reviews/ten24/slatwall/6603)


rmarchandTen24:
@christopher-kundrat ouch we are going to need a migration scritp too

rmarchandTen24:
@christopher-kundrat need updateScript.xml to be updated

PR #6647, opened 2018-08-23
generate Promo codes

rmarchandTen24:


rmarchandTen24:
@ian-a-hickey I may be missing some of the view logic. Will add later

PR #6657, opened 2018-08-30
updating order delivery simple rep

evan1024:


model/entity/OrderDelivery.cfc
```diff
@@ -159,7 +159,7 @@ component displayname="Order Delivery" entityname="SlatwallOrderDelivery" table=
 	// ================== START: Overridden Methods ========================
 
 	public string function getSimpleRepresentation() {
-		return "Delivery for - " & getOrder().getOrderNumber();
+		return "Order Delivery: Order ##" & getOrder().getOrderNumber();
```
jasperboyd: @evan1024 do you need escaped # in cfc?

PR #6680, opened 2018-09-21
void transaction feature 

kangnam:


config/scripts/preupdate/v4_5.cfm
```diff
@@ -26,7 +26,7 @@
 	    </cfquery>
 	<cfelseif ListFind(this.ormSettings.dialect, 'MySQL')>
 	    <cfquery name="local.updateSwProductListingPage" datasource="#this.datasource.name#">
-	        UPDATE SwProductListingPage SET productListingPageID=(SELECT md5(UUID()))
+	        UPDATE SwProductListingPage SET productListingPageID=REPLACE(CAST(UUID() as char character set utf8), '-', '')
```
rmarchandTen24: @kangnam @ian-a-hickey  not sure i understand this change. what was the issue? both make 32 character strings.

migueltarga: The existing line will try to set all productListingPageID to a single UUID and the fix will generate one UUID per record and prevent constraints issues.
this issue only happens on some collations

PR #6684, opened 2018-09-21
delete primary payment method

kangnam:


model/entity/AccountPaymentMethod.cfc
```diff
@@ -124,6 +124,14 @@ component displayname="Account Payment Method" entityname="SlatwallAccountPaymen
 			'12'
 		];
 	}
+	
+	public boolean function isPrimaryPaymentMethod(){
+		return getAccountPaymentMethodID() == getAccount().getPrimaryPaymentMethod().getAccountPaymentMethodID;
```
rmarchandTen24: @kangnam @ian-a-hickey these functions don't call the method.
getAccount().getPrimaryPaymentMethod().getAccountPaymentMethodID
should be 
getAccount().getPrimaryPaymentMethod().getAccountPaymentMethodID()

rmarchandTen24: also they will blow up if no primary is set so you may want to do a null check

model/entity/AccountPaymentMethod.cfc
```diff
@@ -124,6 +124,14 @@ component displayname="Account Payment Method" entityname="SlatwallAccountPaymen
...
+	}
+	
+	public boolean function isNotPrimaryPaymentMethod(){
+		return getAccountPaymentMethodID() == getAccount().getPrimaryPaymentMethod().getAccountPaymentMethodID;
```
rmarchandTen24: here too

PR #6688, opened 2018-09-25
Entity Queue - REVIEW ONLY

migueltarga:


org/Hibachi/HibachiDAO.cfc
```diff
@@ -123,114 +123,19 @@
...
-						}catch(any e){
-							getHibachiScope().logHibachi(e.detail ,true);
-						}
-			    	}
-				}
-		    }
+			// Loop over the modifiedEntities to add updateCalculatedProperties to entity queue
+	    	for(var entity in modifiedEntities){
+	    		if(getService('HibachiService').getEntityHasCalculatedPropertiesByEntityName(entity.getClassName())){
+	    			getHibachiScope().addEntityQueue(entity.getPrimaryIDValue(), entity.getClassName(), 'process#entity.getClassName()#_updateCalculatedProperties');
```
migueltarga: change to addEntityQueueData

org/Hibachi/HibachiScope.cfc
```diff
@@ -423,4 +424,53 @@ component output="false" accessors="true" extends="HibachiTransient" {
 	public boolean function authenticateCollectionPropertyIdentifier(required string crudType, required any collection, required string propertyIdentifier){
 		return getHibachiAuthenticationService().authenticateCollectionPropertyIdentifierCrudByAccount( crudType=arguments.crudType, collection=arguments.collection, propertyIdentifier=arguments.propertyIdentifier, account=getAccount() );
 	}
+	
+	//change name to entityQueueArray
```
migueltarga: remove this comment

org/Hibachi/HibachiDAO.cfc
```diff
@@ -123,114 +123,19 @@
 	    public void function flushORMSession(boolean runCalculatedPropertiesAgain=false) {
 	    	// Initate the first flush
 	    	ormFlush();
+	    	// flush again to persist any changes done during ORM Event handler
+			ormFlush();
```
rmarchandTen24: @migueltarga what is an example of when this becomes an issue? you are referring to the preinsert, preupdate, postinsert,postupdate?

org/Hibachi/HibachiEntityQueueService.cfc
```diff
@@ -1,22 +1,139 @@
...
 
-		//call process from Queued Entity
-		var entityService = getServiceByEntityName( entityName=arguments.entityQueue.getBaseObject() );
-		var processContext = arguments.entityQueue.getProcessMethod();
-		var entity = entityService.invokeMethod( "get#arguments.entityQueue.getBaseObject()#", arguments.entityQueue.getBaseID() );
-		var processData = { '1' = arguments.entity };
-
-		if( arguments.entity.hasProcessObject( processContext ) ){
-			processData['2'] = arguments.entity.getProcessObject( processContext );
+	public void function addQueueHistory(required any entityQueue, required boolean success){
```
rmarchandTen24: @migueltarga I would recommend using sql when it's a bulk process

rmarchandTen24:
@migueltarga I think it's ok to use orm to process a single item however when doing it in bulk we should be using sql otherwise this could impact the systems performance overall

org/Hibachi/Hibachi.cfc
```diff
@@ -1009,12 +1015,37 @@ component extends="framework.one" {
...
+		
+		
+		//Process request entity queue
+		var entityQueueData = getHibachiScope().getEntityQueueData();
+		var entityQueueDataLength = structCount(entityQueueData);
+		if(entityQueueDataLength > 0){
+			
+			var entityQueueArray = [];
+			var currentIndex = 1;
+			var limit = getHibachiScope().setting('globalAllowedOutsideRedirectSites');
```
rmarchandTen24: @migueltarga  should this be globalAllowedOutsideRedirectSites or the processCount setting?

migueltarga: Opps, good one! wrong copy n paste!

org/Hibachi/HibachiDAO.cfc
migueltarga: I reverted to the original flushORMSession function, and you are correct we flush it again to handle the preinsert, preupdate, postinsert,postupdate

org/Hibachi/HibachiEntityQueueService.cfc
migueltarga: we are only using the queue history when it is processing a single EntityQueue object, otherwise, it relies on Workflow history for now.

org/Hibachi/HibachiDAO.cfc
rmarchandTen24: @migueltarga  can you explain a detailed scenario where this is needed because the ormflush commits orm event changes and this should be unnecessary

sumitverma: this is related to when we do additional processing and make changes to other entity from orm event handler. It's been there since early version of Slatwall and is not new.

rmarchandTen24: Qq if the queue batch setting is 0. Then will calculated properties be calculated at all since it looks calculated props are only run via the queue?

rmarchandTen24: is that what entityqueue at the hibachiscope level is supposed to be for? like the request based queue vs the db based one?

migueltarga: YEs, entityqueuedata is all the EntityQueueItems in the request, to prevent us from adding to the db and after getting back from the db..

org/Hibachi/Hibachi.cfc
```diff
@@ -695,6 +698,9 @@ component extends="framework.one" {
                     if(!coreBF.containsBean("hibachiJsonService")){
 						coreBF.declareBean("hibachiJsonService", "#variables.framework.applicationKey#.org.Hibachi.HibachiJsonService",true);
 					}
+					if(!coreBF.containsBean("hibachiEntityQueueDAO")) {
+						coreBF.declareBean("hibachiEntityQueueDAO", "#variables.framework.applicationKey#.org.Hibachi.hibachiEntityQueueDAO", true);
```
rmarchandTen24: component needs proper casing

org/Hibachi/HibachiService.cfc
```diff
@@ -432,7 +432,11 @@
 			} else if ( lCaseMissingMethodName.startsWith( 'export' ) ) {
 				return onMissingExportMethod( missingMethodName, missingMethodArguments );
 			} else if ( lCaseMissingMethodName.startsWith( 'process' ) ) {
-				return onMissingProcessMethod( missingMethodName, missingMethodArguments );
+				if(right(lCaseMissingMethodName,27) == "_updateCalculatedProperties") {
+					return onMissingUpdateCalculatedProperties(missingMethodName, missingMethodArguments)
```
rmarchandTen24: semicolon

PR #6697, opened 2018-10-01
 Include returned orders items in tax calculation report

UltimoTG:
- Returned order items do not have orderfulfillments so doing an `INNER JOIN` on the table excludes them from calculations - changed that to an `LEFT JOIN`.


rmarchandTen24:
https://github.com/ten24/slatwall/pull/6660

PR #6727, opened 2018-10-09
entity queue and calculated properties performance

tdhatcher:
Please review and run locally to test to verify Workflows functionality works as expected. 

I assume a workflow must manually be configured first to process queue if the global setting 'globalEntityQueueDataProcessCount' = 0 otherwise nothing is going to execute for the entity queue processing.

Some of the merging conflicts with the WorkflowService.cfc and HibachiEntityQueueDAO.cfc were a bit less straightforward.

But this should be everything.

### DIFF
--- a/org/Hibachi/HibachiService.cfc
+++ b/org/Hibachi/HibachiService.cfc
@@ -432,7 +432,11 @@
 			} else if ( lCaseMissingMethodName.startsWith( 'export' ) ) {
 				return onMissingExportMethod( missingMethodName, missingMethodArguments );
 			} else if ( lCaseMissingMethodName.startsWith( 'process' ) ) {
-				return onMissingProcessMethod( missingMethodName, missingMethodArguments );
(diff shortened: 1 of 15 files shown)


org/Hibachi/HibachiService.cfc
```diff
@@ -432,7 +432,11 @@
 			} else if ( lCaseMissingMethodName.startsWith( 'export' ) ) {
 				return onMissingExportMethod( missingMethodName, missingMethodArguments );
 			} else if ( lCaseMissingMethodName.startsWith( 'process' ) ) {
-				return onMissingProcessMethod( missingMethodName, missingMethodArguments );
+				if(right(lCaseMissingMethodName,27) == "_updateCalculatedProperties") {
+					return onMissingUpdateCalculatedProperties(missingMethodName, missingMethodArguments)
```
rmarchandTen24: semi colon

PR #6776, opened 2018-10-31
Improved HibachiValidationService Improvements

tdhatcher:
Improved HibachiValidationService to allow validation on deeply nested property within single validation file.


rmarchandTen24:
@tdhatcher this improvement makes sense. However we run into an issue with it because collections overrides hasPropertyByPropertyIdentifier. This prevents reading meta data on the collection itself and instead reads metadata on the collectionObject. When creating a new collection it errors because there is no collectionobject.

PR #6796, opened 2018-11-16
Adding custom inlist delimiter logic

pjohn1024:


model/entity/Collection.cfc
```diff
@@ -3699,7 +3709,7 @@ component displayname="Collection" entityname="SlatwallCollection" table="SwColl
 		if(len(arguments.comparisonOperator)){
 			key &= ":#arguments.comparisonOperator#";
 		}
-		var delimiter = ',';
+		var delimiter = this.getInlistDelimiter();
```
migueltarga: Can you use other delimiters on IN filter in HQL?

model/entity/Collection.cfc
```diff
@@ -2773,7 +2783,7 @@ component displayname="Collection" entityname="SlatwallCollection" table="SwColl
 			predicate = filter.value;
 		}else if(arguments.filter.comparisonOperator eq 'in' || arguments.filter.comparisonOperator eq 'not in'){
 			if(len(filter.value)){
-				predicate = "(" & ListQualify(filter.value,"'") & ")";
+				predicate = "(" & ListQualify(filter.value,variables.inlistDelimiter) & ")";
```
migueltarga: this doesn't look right, the delimiter should be the 3rd parameter

model/entity/Collection.cfc
pjohn1024: @migueltarga I think we are not changing the delimiter on the IN Filter in HQL, The new delimiter is Just used for parsing the in filter list in the predicate

model/entity/Collection.cfc
pjohn1024: Fixed

PR #6803, opened 2018-11-26
migration to use system folder for mounting

rmarchandTen24:
This change is [Reviewable](https://reviewable.io/reviews/ten24/slatwall/6803)

### DIFF
--- a/org/Hibachi/Hibachi.cfc
+++ b/org/Hibachi/Hibachi.cfc
@@ -17,9 +17,6 @@ component extends="framework.one" {
 	// Allow For Instance Config
 	try{include "../../custom/config/configApplication.cfm";}catch(any e){}
 	
-	// Allow For Instance Config
-	try{include "../../custom/system/configApplication.cfm";}catch(any e){}
-	
 	// Allow For DevOps Config
 	try{include "../../../configApplication.cfm";}catch(any e){} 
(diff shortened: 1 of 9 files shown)


org/Hibachi/Hibachi.cfc
```diff
@@ -16,6 +16,10 @@ component extends="framework.one" {
 	try{include "../../config/configApplication.cfm";}catch(any e){}
 	// Allow For Instance Config
 	try{include "../../custom/config/configApplication.cfm";}catch(any e){}
+	
+	// Allow For Instance Config
+	try{include "../../custom/system/configApplication.cfm";}catch(any e){}
```
sumitverma: We don't need to move cfm files, only system generated files should go inn system folder.

org/Hibachi/Hibachi.cfc
```diff
@@ -238,6 +268,21 @@ component extends="framework.one" {
 
     }
     
+    public void function migrateCustomConfigToSystem(){
+    	var systemDirectoryPath = "#this.mappings[ '/#variables.framework.applicationKey#' ]#/custom/system";;
+
```
sumitverma: You will need to create system folder if it doens't exists

rmarchandTen24: @sumitverma all set

PR #6815, opened 2018-12-04
Location creation fix

gerickson808:


model/entity/Location.cfc
```diff
@@ -141,7 +141,7 @@ component displayname="Location" entityname="SlatwallLocation" table="SwLocation
 			
 			//Add each of the parents in the chain to the string.
 			var parentLocation = this.getParentLocation();
-			while (!isNull(parentLocation)){
+			while (!isNull(parentLocation) && !isStruct(parentLocation)){
```
rmarchandTen24: @gerickson808 shouldn't we handle the data if it is a struct?

PR #6822, opened 2018-12-07
only run inventory logic if we're configured to track inventory

jasperboyd:


model/service/SkuService.cfc
```diff
@@ -617,6 +617,8 @@ component extends="HibachiService" persistent="false" accessors="true" output="f
 
 	public any function processSku_updateInventoryCalculationsForLocations(required any sku) {
 		
+		if(getSettingService().getSettingValue('skuTrackInventoryFlag')){
```
rmarchandTen24: shouldn't this be arguments.sku.setting('skuTrackInventoryFlag')

jasperboyd: @rmarchandTen24 yes thank you good catch

jasperboyd: @rmarchandTen24 resolved sorry bout the whitespace

PR #6834, opened 2018-12-17
Develop pj portto core

pjohn1024:


rmarchandTen24:
@pjohn1024 is there a view aspect to this at all?

PR #6851, opened 2019-01-03
Use primary ID as group by if exists in columns & hibachi listing tag improvement

jasperboyd:


model/entity/Collection.cfc
```diff
@@ -3336,7 +3336,9 @@ component displayname="Collection" entityname="SlatwallCollection" table="SwColl
 						) continue;
 						
 						if(getService('HibachiService').getPrimaryIDPropertyNameByEntityName(getCollectionObject()) == convertALiasToPropertyIdentifier(column.propertyIdentifier)){
-							groupByList = listprepend(groupByList, column.propertyIdentifier);
+							//if we have the collection objects primary id property as a column exclude all others group bys for better performance
```
rmarchandTen24: @jasperboyd looks good but can you use the deleted behavior if it's a report

jasperboyd: @rmarchandTen24 it will only get to this logic if it isn't a report L3272

PR #6863, opened 2019-01-08
resolved the conflicts

mfsi-nirmitm:
resolved the conflicts


org/Hibachi/client/src/alert/service/alertservice.ts
```diff
@@ -22,26 +22,13 @@ export class AlertService {
...
         this.alerts.push(alert);
         setTimeout(()=> {
             if(!alert.dismissable){
                 this.removeAlert(alert);
             }
         }, 3500);
-=======
-    addAlert = (alert:any):void =>{
-        
-        if(alert.msg){
```
rmarchandTen24: @mfsi-nirmitm we should be incorporating the new changes rather than dismissing them

PR #6885, opened 2019-01-18
fixing issue where all skus got the same image in price manager

evan1024:


admin/client/src/sku/components/imagedetailmodallauncher.html
```diff
@@ -42,7 +42,7 @@
                     <div class="row">
                         <div class="col-md-6">
                             <div class="form-group" ng-if="!swImageDetailModalLauncher.customImageNameFlag" id="j-standard-image-name">
```
rmarchandTen24: @evan1024 does the customImageNameFlag do anything?

admin/client/src/sku/components/swimagedetailmodallauncher.ts
```diff
@@ -107,16 +107,17 @@ class SWImageDetailModalLauncherController{
         data.append('ajaxRequest', "1");
 
         data.append('productID', this.swPricingManager.productId);
-
-        if(this.customImageNameFlag){
```
rmarchandTen24: @evan1024 it's like it was intended to be here for something

admin/client/src/sku/components/imagedetailmodallauncher.html
evan1024: It controls whether or not it will show the input field for imageFileName.

admin/client/src/sku/components/swimagedetailmodallauncher.ts
evan1024: It controls whether or not the input for imageFileName shows up in the view. However, if the flag was true and you didn't type anything into imageFileName, the append would throw an error. This checks for the actual existence of the value.

admin/client/src/sku/components/imagedetailmodallauncher.html
rmarchandTen24: @evan1024 there is no binding for it though

evan1024: https://github.com/ten24/slatwall/pull/6885/files#diff-ad39edae9c7eb02a748e67535540d502R53

PR #6892, opened 2019-01-18
ability to set org flag

evan1024:


rmarchandTen24:
@evan1024 does this have a model aspect to it?
I see it's at the order level and I know we have it at the account level in core

evan1024:
> @evan1024 does this have a model aspect to it?
> I see it's at the order level and I know we have it at the account level in core

It delegates those values to account create https://github.com/ten24/slatwall/blob/develop/model/service/OrderService.cfc#L1134

Setting the org flag on the account level.
https://github.com/ten24/slatwall/blob/develop/model/entity/Account.cfc#L62

PR #6903, opened 2019-01-18
removing condition that no longer exists, which prevents primary email address check

jasperboyd:


jasperboyd:
Closing to make some additional changes

PR #6905, opened 2019-01-18
readding missing account validation condition

jasperboyd:


rmarchandTen24:
@jasperboyd it looks like slatwallAuthenticationExistsFlag doesn't exist in hotfix. check circle

PR #6906, opened 2019-01-21
Bringing in changes from ecom

pjohn1024:


admin/views/entity/accounttabs/cartsandquotes.cfm
```diff
@@ -74,6 +71,7 @@ Notes:
 
 <hb:HibachiListingDisplay collectionList="#rc.ordersNotPlacedCollectionList#"
 	recordEditAction="admin:entity.editorder"
-	recordDetailAction="admin:entity.detailorder"
+	recordDetailAction="admin:entity.detailorder" 
+	usingPersonalCollection="true"
```
rmarchandTen24: @pjohn1024 @christopher-kundrat I believe Chris made a change where this should be false

PR #6938, opened 2019-01-31
use plural rbkeys for aggregates on entities

rmarchandTen24:
This change is [Reviewable](https://reviewable.io/reviews/ten24/slatwall/6938)


org/Hibachi/client/src/collection/components/swdisplayoptions.ts
```diff
@@ -94,12 +94,19 @@ class SWDisplayOptions{
...
                     var currentEntity;
                     var currentEntityInstance;
                     var prefix = 'entity.';
                     if(selectedProperty.$$group == "attribute"){
                         return selectedProperty.displayPropertyIdentifier;
                     }
+                    //if is aggregate of an object
+                    if(selectedProperty.aggregate && selectedProperty.cfc){
+                        var lastEntityName = $hibachi.getLastEntityNameInPropertyIdentifier(baseEntityCfcName,actualPropertyIdentifier);
+                        title = rbkeyService.getRBKey(prefix+lastEntityName+'_plural');
```
migueltarga: I think we need to add the aggregate function,
in case we have SUM and AVG for example added, the titles should be **Orders SUM**, **Orders AVG**.

rmarchandTen24: I don't think you can have an orders sum or avg. You would have a sum or avg on property but not the object

christopher-kundrat: what about something like Orders Count?

PR #6956, opened 2019-02-06
Removed 'Required Details' label in place order module

samanthatran:


admin/views/entity/preprocessorder_placeorder.cfm
```diff
@@ -64,8 +64,6 @@ Notes:
 		<hb:HibachiPropertyRow>
 			<hb:HibachiPropertyList>
 
-				<h3>Required Details</h3>
```
rmarchandTen24: @samanthatran I don't think we should remove this. However we should only show it when we get something back from the orders requirement list.
you could add a cfif on len(rc.order.getOrderRequirementsList())

then it will only show if an item is in the list

samanthatran: Thanks for your input Ryan, I've just added the update

PR #6972, opened 2019-02-14
skipping filters with no results

evan1024:
skipping filters with no results


integrationServices/slatwallcms/skeletonsite/tags/tagtemplates/FilterCountDisplayItem.cfm
```diff
@@ -32,6 +32,9 @@
 		   	<div id="collapse#uuid#" class="collapse" aria-labelledby="#uuid#FilterCountDisplayItem" data-parent="##accordion">
 				<div class="card-body">
             		<cfloop array="#arguments.optionData#" index="option">
+			        <cfif structKeyExists(option,'COUNT') AND option['COUNT'] EQ 0 >
```
rmarchandTen24: @evan1024 what if someone had the filter applied?

rmarchandTen24: @evan1024 there is code about whether or not the options was selected. this would prevent being able to unselect the option

PR #6977, opened 2019-02-19
making sure db data runs alphabetically

evan1024:


org/Hibachi/HibachiDataService.cfc
```diff
@@ -124,6 +124,7 @@ component accessors="true" output="false" extends="HibachiService" {
 
 	public boolean function loadDataFromXMLDirectory(required string xmlDirectory, boolean ignorePreviouslyInserted=true) {
 		var dirList = directoryList(arguments.xmlDirectory);
+		arraySort(dirList,"textnocase");
```
rmarchandTen24: @evan1024 if you are having an issue with ordering then you can declare a dependency via this https://github.com/ten24/slatwall/blob/develop/config/dbdata/SlatwallWorkflowTask.xml.cfm#L2

PR #6993, opened 2019-02-28
adding null checks for address zone and zone location

kangnam:


kangnam:
https://ten24.teamwork.com/index.cfm#tasks/23524055

PR #6997, opened 2019-03-01
added Emails Tab (displays if entity has an email template)

samanthatran:


org/Hibachi/HibachiTags/HibachiEntityDetailGroup.cfm
```diff
@@ -117,6 +117,55 @@
...
+											    data-has-search="true"
+											    record-edit-action="admin:entity.editemail"
+											    record-edit-query-string="redirectAction=admin:entity.listemail"
+											    record-detail-action="admin:entity.detailemail"
+											    data-is-angular-route="false"
+											    data-angular-links="false"
+											    data-has-action-bar="false"
+											    data-multi-slot="true"
+											>
+											    <sw-listing-column data-property-identifier="emailID" data-is-visible="false" data-is-deletable="false" ></sw-listing-column>
```
rmarchandTen24: @samanthatran we shouldn't be using sw-listing-column. It is deprecated because it doesn't migrate to angular 2 since it uses transclusion.

PR #7019, opened 2019-03-27
Collection Group By Refactor

migueltarga:
- if AddGroupBy() function is used, it takes precedence.
- if x-to-many filter is added to Collection, automatically group by to prevent dupes
- if  x-to-many display property added, duplicate records
- if x-to-many display property and filter added, duplicate records
- if using aggregate, automatically group by

With this code, we can support add *-to-many columns on listing display not just aggregates


meta/tests/unit/entity/CollectionTest.cfc
```diff
@@ -664,11 +739,331 @@ component extends="Slatwall.meta.tests.unit.entity.SlatwallEntityTestBase" {
...
+		assertTrue(arrayLen(pageRecords) == recordsCount, "Wrong amount of products returned! Expecting #recordsCount# records but returned #arrayLen(pageRecords)#");
+		
 		assertTrue(arrayLen(pageRecords) == 1, "Wrong amount of products returned! Expecting 1 record but returned #arrayLen(pageRecords)#");
 	}
+	
+	
+	/**
+	* @test
+	*/
+	public void function oneToManyPropertyShouldDuplicateRecordsTest(){
```
rmarchandTen24: @migueltarga it looks like some of these tests are duplicated function names. Are they copy and paste, or should I assume they need the names updated?

PR #7027, opened 2019-03-29
added attribute options count bubble to Attribute Detail

samanthatran:


admin/views/entity/detailattribute.cfm
```diff
@@ -79,7 +79,7 @@ Notes:
 			<hb:HibachiEntityDetailGroup object="#rc.attribute#">
 				<hb:HibachiEntityDetailItem view="admin:entity/attributetabs/basic" open="true" text="#$.slatwall.rbKey('admin.define.basic')#" showOnCreateFlag=true />
 				<cfif not rc.attribute.getNewFlag() and listFindNoCase( "text,password,checkboxGroup,multiselect,radioGroup,select",rc.attribute.getAttributeInputType() )>
-					<hb:HibachiEntityDetailItem view="admin:entity/attributetabs/attributeoptions" />
+					<hb:HibachiEntityDetailItem count="#rc.attribute.getAttributeOptionsSmartlist().getRecordsCount()#" view="admin:entity/attributetabs/attributeoptions" />
```
rmarchandTen24: @samanthatran can you change this to use attributeOptionsCollectionlist instead of smartlist?

migueltarga: @rmarchandTen24 `getAttributeOptionsCount()` should work, right?

rmarchandTen24: @migueltarga i think so

PR #7032, opened 2019-04-02
Filter max-min range set for numbers.

krishnamac05:
added min-max input value for number range filter.

### DIFF
--- a/org/Hibachi/client/src/collection/components/sweditfilteritem.ts
+++ b/org/Hibachi/client/src/collection/components/sweditfilteritem.ts
@@ -311,10 +311,23 @@ class SWEditFilterItem{
                                     filterItem.attributeID = selectedFilterProperty.attributeID;
                                     filterItem.attributeSetObject = selectedFilterProperty.attributeSetObject;
                                 }
-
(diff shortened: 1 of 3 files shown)


org/Hibachi/client/src/collection/components/sweditfilteritem.ts
```diff
@@ -311,10 +311,23 @@ class SWEditFilterItem{
                                     filterItem.attributeID = selectedFilterProperty.attributeID;
                                     filterItem.attributeSetObject = selectedFilterProperty.attributeSetObject;
                                 }
-
+                                //scope.inListArray = [];
                                 filterItem.comparisonOperator = selectedFilterProperty.selectedCriteriaType.comparisonOperator;
-
-                                //retrieving implied value or user input | ex. implied:prop is null, user input:prop = "Name"
+                                
+                                if(scope.filterItem.comparisonOperator === 'between' || scope.filterItem.comparisonOperator === 'not between'){
```
rmarchandTen24: @krishnamac05 this code should be specific to the criteria number component

PR #7066, opened 2019-05-21
Updating arguments for test domain

christopher-kundrat:
This change is [Reviewable](https://reviewable.io/reviews/ten24/slatwall/7066)


model/service/EmailService.cfc
```diff
@@ -63,7 +63,7 @@ Notes:
 		<cfset var whiteListedDomains = getHibachiScope().setting('globalWhiteListedEmailDomains') />
 		<cfset var emailTestDomain = getHibachiScope().setting('globalTestingEmailDomain') />
 		<cfloop list="#arguments.emailAddresses#" index="local.emailAddress">
-			<cfif listFindNoCase(listLast(local.emailAddress, '@'), whiteListedDomains)>
+			<cfif listFindNoCase(whiteListedDomains, listLast(local.emailAddress, '@'))>
```
migueltarga: good catch, it works if the setting has a single domain, that as the false positive during tests

